PR #2, opened 2018-03-07
Echo server

parkerpersaud:
https://trello.com/c/jGGLW7TK/68-925-echo-server


README.md
```diff
@@ -25,8 +25,11 @@ There are three different types of tagged images:
 that do not change.  
 
 ### Run project 
+The default port is set to 4000 but can be set to specified port by passing a port number as an 
+argument.
+
 ```
-$ docker run klkelley/http-server:latest
+$ docker run klkelley/http-server:master.latest <OPTIONAL PORT NUMBER>
```
aaronlahey: For optional values I'd like to pass it as `-p 8080` or `--port 8080`. I'd also like the default port to be `0`.

src/main/java/me/karakelley/http/EchoHandler.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.PrintWriter;
+import java.net.Socket;
+
+class EchoHandler extends Thread {
+  private final Socket clientSocket;
+  private final Logger logger = LoggerFactory.getLogger(EchoHandler.class.getName());
```
aaronlahey: Generally `Logger`s from these libraries are thread-safe. You can make this `private static final`. I'm also surprised to see `getName`, Can you just pass the `Class` instance into `getLogger`?

src/main/java/me/karakelley/http/EchoHandler.java
```diff
@@ -0,0 +1,42 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.PrintWriter;
+import java.net.Socket;
+
+class EchoHandler extends Thread {
```
aaronlahey: Why extend `Thread`?

src/main/java/me/karakelley/http/EchoHandler.java
```diff
@@ -0,0 +1,42 @@
...
+      logger.info(e.getMessage());
+    }
+  }
+
+  private void echoBack(BufferedReader in, PrintWriter out) throws IOException {
+    String input;
+    while ((input = in.readLine()) != null) {
+      out.println("echo: " + input);
+    }
+    closeConnections(in, out);
```
aaronlahey: Read up on `try-with-resources`. I don't think you need to close these manually.

src/main/java/me/karakelley/http/EchoHandler.java
aaronlahey: I don't think you need to name everything with the word `Echo`. The "echo" functionality is just a stepping stone.

src/test/java/me/karakelley/http/EchoServerTest.java
```diff
@@ -0,0 +1,107 @@
...
+  void tearDown() throws IOException {
+    client.terminateConnection();
+  }
+
+  @Test
+  void testEchoesBackGivenAPort() throws IOException, InterruptedException {
+    server = new EchoServer(new String[]{"5000"}, logger);
+    ExecutorService executorService = Executors.newFixedThreadPool(1);
+    executorService.submit(() -> server.start());
+    Thread.sleep(10);
```
aaronlahey: EEEEEK. Sleeping in a test is definitely a smell. Why do you need to do this? You're going to slow down every test to 10ms and also, potentially have flaky tests if 10ms is just *barely* long enough.

src/test/java/me/karakelley/http/EchoServerTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+class EchoServerTest {
+  private EchoClient client;
+  private EchoServer server;
+  private EchoServer serverOnSamePort;
+  private Logger logger;
+
+  @BeforeEach
+  void setUp() throws InterruptedException {
+    logger = mock(Logger.class);
```
aaronlahey: Let's not use mockito in this project. Try rolling your own fakes.

src/main/java/me/karakelley/http/EchoServer.java
```diff
@@ -0,0 +1,57 @@
...
+      try {
+        serverSocket.close();
+      } catch (IOException e) {
+        logger.info(e.getMessage());
+      }
+    }
+  }
+
+  private Integer setPort(String[] args) {
+    return (args.length != 1 || invalidPort(args)) ? DEFAULT_PORT : Integer.parseInt(args[0]);
```
aaronlahey: If I pass two arguments or the port is invalid you're going to use the default port? That's surprising behavior. I'd expect it to bail and log an error.

src/test/java/me/karakelley/http/EchoServerTest.java
```diff
@@ -0,0 +1,107 @@
...
+    client = new EchoClient();
+    serverOnSamePort = new EchoServer(new String[]{"4050"}, logger);
+    ExecutorService executorService = Executors.newFixedThreadPool(1);
```
aaronlahey: There's a more appropriate `ExecutorService` to use here.

aaronlahey: Why are you setting up and running two servers for every test?

src/test/java/me/karakelley/http/EchoServerTest.java
```diff
@@ -0,0 +1,107 @@
...
+    Thread.sleep(10);
+  }
+
+  @AfterEach
+  void tearDown() throws IOException {
+    client.terminateConnection();
+  }
+
+  @Test
+  void testEchoesBackGivenAPort() throws IOException, InterruptedException {
```
aaronlahey: Often times in tests I just default to declaring `throws Exception` so I don't need to enumerate all the possible checked exceptions that can be thrown. Sometimes I'll just take whatever method throws a checked exception, catch the exception, and wrap it in a `RuntimeException`...

```
try {
  socket.close();
} catch (IOException e) {
  throw new RuntimeException(e);
}
```

Anything you can do to reduce noise in your tests is typically a good thing.

aaronlahey (changes requested):

src/test/java/me/karakelley/http/EchoServerTest.java
andarcabrera: +1

src/main/java/me/karakelley/http/EchoServer.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  private boolean invalidPort(String[] args) {
```
andarcabrera: The port validation can be extracted from this class. It seems more setup-ish.

PR #3, opened 2018-03-09
Change log statement to state port chosen by system for port 0

parkerpersaud:
https://trello.com/c/HyVjFrT9/85-1-ask-socket-for-correct-port-number

### DIFF
--- a/src/main/java/me/karakelley/http/Server.java
+++ b/src/main/java/me/karakelley/http/Server.java
@@ -3,6 +3,8 @@ package me.karakelley.http;
 import java.io.*;
 import java.net.ServerSocket;
 
+import static java.lang.String.format;
+
 class Server {
   private final int port;
   private ServerSocket serverSocket;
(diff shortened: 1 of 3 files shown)


andarcabrera (approved):

src/main/java/me/karakelley/http/Server.java
```diff
@@ -16,7 +16,7 @@ public Server(int port, ServerLogger logger) {
   public void start() {
     try {
       serverSocket = new ServerSocket(port);
-      logger.info("Started on port " + port);
+      logger.info("Started on port " + serverSocket.getLocalPort());
```
aaronlahey, line 21: Don't hate me but I have two comments on this one line:

1. Aren't you testing logging? Why is there no corresponding test change?
2. This is not the ideal way to interpolate values when using a logging framework.

aaronlahey (changes requested):

PR #6, opened 2018-03-13
Feat get

parkerpersaud:
https://trello.com/c/wStZbilY/69-950-get-request


src/main/java/me/karakelley/http/Request.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.io.BufferedReader;
+import java.util.HashMap;
+
+public class Request {
+  private final RequestLine requestLine;
+  private final RequestHeaders requestHeaders;
+  private final BufferedReader reader;
+
+  public Request(BufferedReader reader) {
```
aaronlahey: Why depend on such a concrete class? Would `InputStream` suffice?

src/main/java/me/karakelley/http/HttpServer.java
```diff
@@ -0,0 +1,37 @@
...
+  private Integer portNumber;
+
+  public HttpServer(int port, ServerLogger logger) {
+    this.port = port;
+    this.logger = logger;
+  }
+
+  public void start() {
+    try (ServerSocket serverSocket = new ServerSocket(port)) {
+      setPortNumber(serverSocket);
```
aaronlahey: Any time you can avoid state like this is a win. It makes the code harder to reason about. The only interface your `HttpServer` needs to adhere to is `getPortNumber`. Why not just ask the server what the local port is in that getter (that's the beauty of getters if there is one).

src/main/java/me/karakelley/http/HttpServer.java
```diff
@@ -0,0 +1,37 @@
...
+      logger.info("Started on port " + portNumber);
+      while (true)
+        new HttpConnectionHandler(serverSocket.accept()).start();
+    } catch (IOException | IllegalArgumentException e) {
```
aaronlahey: What would cause both of these exceptions?

src/main/java/me/karakelley/http/Request.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public String getPath() {
+    return requestLine.getPath();
+  }
+
+  public String getProtocol() {
+    return requestLine.getProtocol();
+  }
+
+  public HashMap getHeaders() {
```
aaronlahey: 1. Why commit to a concrete implementation here?
2. Any specific reason for not specifying the types of the keys/values?

src/main/java/me/karakelley/http/Request.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.HashMap;
+
+public class Request {
+  private final RequestLine requestLine;
+  private final RequestHeaders requestHeaders;
+  private final BufferedReader reader;
+
+  public Request(BufferedReader reader) {
+    this.reader = reader;
+    this.requestLine = setRequestLine();
```
aaronlahey: It doesn't look like this method is actually setting anything but its name suggests otherwise.

src/main/java/me/karakelley/http/Request.java
```diff
@@ -0,0 +1,62 @@
...
+    this.requestHeaders = setHeaders();
```
aaronlahey: Same comment here about naming.

src/main/java/me/karakelley/http/RequestHeaders.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.HashMap;
+
+public class RequestHeaders {
+  HashMap headers;
+
+  public RequestHeaders(HashMap headers) {
+    this.headers = headers;
+  }
+
+  public HashMap getHeaders() {
```
aaronlahey: 1. `requestHeaders.getHeaders()`? Seems strange.
2. What does this abstraction bring to the table? What's its reason for existing?

src/main/java/me/karakelley/http/Response.java
```diff
@@ -0,0 +1,23 @@
...
+  private final String OK = "200 OK\r\n";
+  private final String BAD = "400 Bad Request\r\n";
+  private final String BLANK_LINE = "\r\n";
+
+  public Response(Request request) {
+    this.request = request;
+  }
+
+  public String getResponse() {
+    if (request.getPath().equals(ROOT)) {
```
aaronlahey: Why return a 400 here if the request is not to `/`?

src/test/java/me/karakelley/http/HttpServerTest.java
```diff
@@ -0,0 +1,169 @@
...
+    maxTries = 4;
+  }
+
+  @Test
+  void testSendsResponseGivenValidRequest() throws Exception {
+    httpServer = new HttpServer(0, mockLogger);
+    ExecutorService executorService = Executors.newSingleThreadExecutor();
+    executorService.submit(() -> httpServer.start());
+
+    while (true) {
```
aaronlahey: Lots of duplication in this test.

aaronlahey (commented):
1. Sending a single `\r\n` in `nc` throws a `java.lang.ArrayIndexOutOfBoundsException`.
1. Sending "GET" throws a `java.lang.ArrayIndexOutOfBoundsException`.
1. ending "GET /" throws a `java.lang.ArrayIndexOutOfBoundsException`.

1. "Clients SHOULD be tolerant in parsing the Status-Line and servers tolerant when parsing the Request-Line. In particular, they SHOULD accept any amount of SP or HT characters between fields, even though only a single SP is required."
    ```
    ~ :: nc localhost 8080
         GET      /     HTTP/1.1
    
    HTTP/1.1 400 Bad Request
    ```
1. Your HTTP version can be anything....

    ```
    ~ :: nc localhost 8080
    GET / H
    
    HTTP/1.1 200 OK
    Content-Type: text/plain
    
    Hello World
    ```

1. None of your exceptions log the stack trace...

    ```
    2018-03-15T00:58:53.679+0000 INFO HttpConnectionHandler [ForkJoinPool.commonPool-worker-0] java.lang.ArrayIndexOutOfBoundsException: 1
    ```

1. Your method can by anything...

    ```
    ~ :: nc localhost 8080
    GIT / HTTP/1.1
    
    HTTP/1.1 200 OK
    Content-Type: text/plain
    
    Hello World
    ```

1. Requests with a body should have a content length:

    ```
    HTTP/1.1 200 OK
    Content-Type: text/plain
    
    Hello World~
    ```

>14.13 Content-Length
The Content-Length entity-header field indicates the size of the entity-body, in decimal number of OCTETs, sent to the recipient or, in the case of the HEAD method, the size of the entity-body that would have been sent had the request been a GET... Applications SHOULD use this field to indicate the transfer-length of the message-body...

src/main/java/me/karakelley/http/HttpConnectionHandler.java
```diff
@@ -0,0 +1,35 @@
...
+import java.net.Socket;
+import java.util.concurrent.CompletableFuture;
+
+class HttpConnectionHandler {
+  private final Socket clientSocket;
+  private static ServerLogger logger;
+
+  public HttpConnectionHandler(Socket clientSocket) {
+    this.clientSocket = clientSocket;
+    this.logger = new ServerLogger("HttpConnectionHandler");
```
andarcabrera: why a string vs `HttpConnectionHandler.class`?

src/main/java/me/karakelley/http/Request.java
parkerpersaud: I was depending on `BufferedReader` in order to use `readLine()`

src/test/java/me/karakelley/http/HttpServerTest.java
klkelley: In regards to initializing the `httpServer` then starting it two lines down?

src/main/java/me/karakelley/http/RequestHeaders.java
parkerpersaud: While it may be an early abstraction it also seems like it can easily be it's own entity

src/main/java/me/karakelley/http/HttpServer.java
parkerpersaud: I was having issues using a getter using try-with-resources because the `serverSocket` was being initialized within the try-with-resources causing a null pointer exception. This was what I could come up with to make it work with try-with-resources. I just changed it to use a getter with a `finally` block instead.

src/main/java/me/karakelley/http/Request.java
andarcabrera: Can you depend on an interface that has `readLine()` instead?

src/main/java/me/karakelley/http/Response.java
andarcabrera: If `request.getPath()` is null for some reason, you can have a null pointer exception. I would use the constant first in the comparison when possible. In your case `ROOT`

src/main/java/me/karakelley/http/HttpConnectionHandler.java
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  public void start() {
+    CompletableFuture.runAsync(() -> {
+      try (LineReader reader = new BufferedLineReader(new InputStreamReader(clientSocket.getInputStream()));
+           PrintWriter out = new PrintWriter(clientSocket.getOutputStream())) {
+        Request newRequest = new Request(reader);
+        sendResponse(newRequest, out);
+      } catch (Exception e) {
+        logger.info(e.getMessage());
```
aaronlahey: Generally when you're logging exceptions it's best to just pass the entire exception...

```
logger.info("Ouch!", e);
```

The logging framework will genearlly take care of logging the message and the entire stack trace for you.

src/main/resources/logback.xml
```diff
@@ -7,9 +7,9 @@
             </pattern>
         </encoder>
     </appender>
-    <appender name="appender" class="me.karakelley.http.Appender"/>
+    <appender name="appender" class="me.karakelley.http.utility.Appender"/>
```
aaronlahey: Does this have to be declared in the logback.xml? Is this specific appender implementation specific to tests?

src/test/java/me/karakelley/http/HttpServerTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+    try {
+      Thread.sleep(10);
+      client.connect("127.0.0.1", httpServer.getPortNumber());
+    } catch (ConnectException ex) {
+      count++;
+      if (count > maxTries) {
+        throw ex;
+      }
+    }
```
aaronlahey: I still feel like there's a lot of duplication in these tests. You've got a nice implementation of `ClientHelper` going here. Maybe there's an opportunity to turn your `connect` method into a `connectWithRetry`.

src/test/resources/logback-test.xml
```diff
@@ -1,7 +1,7 @@
 <configuration>
     <statusListener class="ch.qos.logback.core.status.NopStatusListener" />
 
-    <appender name="appender" class="me.karakelley.http.Appender">
+    <appender name="appender" class="me.karakelley.http.utility.Appender">
```
aaronlahey: I would rename this class to describe what it does. `InMemoryAppender`, `FakeAppender`, `MockAppender`, etc.

aaronlahey: (It's [not really a mock](https://martinfowler.com/bliki/TestDouble.html), though)

src/test/java/me/karakelley/http/RequestTest.java
```diff
@@ -0,0 +1,76 @@
...
+    assertEquals(headers, request.getHeaders());
+  }
+
+  @Test
+  void testOneHeader() {
+    request = new Request(newBufferedReader(basicGetRequest));
+    assertEquals("localhost:5000", request.getHeader("Host"));
+  }
+
+
```
aaronlahey: Extra line.

src/test/java/me/karakelley/http/RequestTest.java
```diff
@@ -0,0 +1,76 @@
...
+  @Test
+  void testBadHeaders() {
+    try {
+      request = new Request(newBufferedReader("GET / HTTP/1.1\r\nHOST=badrequest"));
+    } catch (Exception e) {
+      assertEquals("java.lang.ArrayIndexOutOfBoundsException: 1", e.getMessage());
+    }
+  }
+
+
```
aaronlahey: Extra line.

src/test/java/me/karakelley/http/RequestTest.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.HashMap;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class RequestTest {
+  Request request;
+  final String basicGetRequest = "GET / HTTP/1.1\r\nHost: localhost:5000\r\n";
+  final String requestWithHeaders = "GET / HTTP/1.1\r\nHost: localhost:5000\r\nConnection: keep-alive\r\n";
+
+
```
aaronlahey: Extra line.

aaronlahey (approved):
Left a bit of feedback, but approved otherwise. Kudos for having much more consistent code this project.

src/test/java/me/karakelley/http/HttpServerTest.java
parkerpersaud: Oh that's a great idea!

PR #9, opened 2018-03-20
Feat redirect

parkerpersaud:
https://trello.com/c/8zTo7X9e/83-300-redirect

### DIFF
--- a/src/main/java/me/karakelley/http/Request.java
+++ b/src/main/java/me/karakelley/http/Request.java
@@ -7,13 +7,15 @@ import java.util.Map;
 
 public class Request {
   private final LineReader reader;
+  private final int port;
   private String requestMethod;
   private String requestPath;
   private String requestProtocol;
-  Map headers = new HashMap<String, String>();
+  final Map<String, String> headers = new HashMap<>();
 
-  public Request(LineReader reader) {
+  public Request(LineReader reader, int port) {
     this.reader = reader;
+    this.port = port;
     setRequestLine();
     setHeaders();
   }
@@ -35,7 +37,19 @@ public class Request {
   }
 
   public String getHeader(String header) {
-    return (String) headers.get(header);
+    return headers.get(header);
+  }
+
+  public int getPort() {
+    return port;
+  }
+
+  public boolean validRequestLine() {
+    try {
+      return requestMethod.equals("GET") && requestProtocol.equals("HTTP/1.1");
+    } catch (NullPointerException e) {
+      return false;
+    }
   }
 
   private void setHeaders() {
@@ -63,14 +77,6 @@ public class Request {
       }
     } catch (Exception e) {
       throw new RuntimeException(e);
-    }
-  }
-
-  public boolean validRequestLine() {
-    try {
-      return requestMethod.equals("GET") && requestProtocol.equals("HTTP/1.1");
-    } catch (NullPointerException e) {
-      return false;
     }
   }
 }
--- a/src/main/java/me/karakelley/http/ResponseFormatter.java
+++ b/src/main/java/me/karakelley/http/ResponseFormatter.java
@@ -1,0 +1,30 @@
+package me.karakelley.http;
+
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class ResponseFormatter {
+  private final static String PROTOCOL = "HTTP/1.1 ";
+  private final static String CRLF = "\r\n";
+
+  public static String build(String statusLine, Map<String, String> headers, String body) {
+    StringBuilder response = new StringBuilder();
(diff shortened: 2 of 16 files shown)


src/main/java/me/karakelley/http/Request.java
```diff
@@ -38,6 +40,18 @@ public String getHeader(String header) {
     return (String) headers.get(header);
```
aaronlahey: Why the cast?

aaronlahey (approved):
I have comments, but let's merge this, continue on with your other stories, and schedule some time to talk more thoroughly about where your design ends up on Friday.

src/main/java/me/karakelley/http/Router.java
```diff
@@ -0,0 +1,23 @@
...
+
+  public static Controller dispatch(Request request) {
+    Map<String, Controller> routes = new HashMap();
+    routes.put("/redirectme", new RedirectController(request));
+    routes.put("/", new RootController(request));
+
+    if (routes.get(request.getPath()) == null) {
+      return new InvalidRequestController();
+    }
+    return routes.get(request.getPath());
```
mikeknep, line 21: Style nit: these lines could be simplified with [Map#getOrDefault](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#getOrDefault-java.lang.Object-V-)

src/main/java/me/karakelley/http/ResponseFormatter.java
```diff
@@ -0,0 +1,30 @@
...
+
+  public static String formatHeaders(Map<String, String> headers) {
+    return headers.entrySet()
+            .stream()
+            .map(entry -> entry.getKey() + ": " + entry.getValue() + "\r\n")
+            .collect(Collectors.joining());
+  }
+
+  public static String formatStatusLine(String statusLine) {
+    return PROTOCOL + statusLine + "\r\n";
```
mikeknep, line 28: Any reason this method and `formatHeaders` aren't using `CRLF`?

src/main/java/me/karakelley/http/controllers/RedirectController.java
```diff
@@ -0,0 +1,47 @@
...
+    return response;
+  }
+
+  public Response setResponse(Response response) {
+    response.setStatus(Status.MOVED_PERMANENTLY);
+    response.setHeaders("Location", "http://" + findHostAndPort() + "/");
+    return response;
+  }
+
+  public String findHostAndPort() {
```
mikeknep, line 32: Does this need to be public? It's not in the `Controller` interface, so anything calling this would have to know it has an instance of this concrete class

src/main/java/me/karakelley/http/controllers/RootController.java
```diff
@@ -0,0 +1,27 @@
...
+  private final Request request;
+  private final String DEFAULT_RESPONSE = "Hello World";
+
+  public RootController(Request request) {
+    this.request = request;
+  }
+
+  public Response respond() {
+    Response response = new Response();
+    if (request.validRequestLine() && request.getMethod().equals("GET")) {
```
mikeknep, line 17: Possibly still too early for an abstraction, but this exact logic also exists in the `RedirectController`

src/test/java/me/karakelley/http/HttpServerTest.java
```diff
@@ -161,4 +163,40 @@ void testSocketNotAvailable() {
       assertEquals("Connection refused (Connection refused)", ex.getMessage());
     }
   }
+
+  @Test
+  void testRedirectsWhenPathIsRedirectme() throws Exception {
+    httpServer = new HttpServer(4000);
+    ExecutorService executorService = Executors.newSingleThreadExecutor();
+    executorService.submit(() -> httpServer.start());
+    Thread.sleep(10);
```
mikeknep, line 172: Why are these `sleep`s necessary?

src/main/java/me/karakelley/http/controllers/InvalidRequestController.java
```diff
@@ -0,0 +1,13 @@
+package me.karakelley.http.controllers;
+
+import me.karakelley.http.Response;
+import me.karakelley.http.Status;
+
+public class InvalidRequestController implements Controller {
+
+  public Response respond() {
+    Response response = new Response();
+    response.setStatus(Status.NOT_FOUND);
```
mikeknep, line 10: `404` vs `400`?

src/main/java/me/karakelley/http/Router.java
parkerpersaud: That's awesome - hadn't seen that before!

PR #10, opened 2018-03-22
Feat list files

parkerpersaud:
https://trello.com/c/hi7JgE3g/73-550-pass-a-directory-path-at-the-command-line-and-list-all-files-at-root-path


src/main/java/me/karakelley/http/HttpConnectionHandler.java
```diff
@@ -32,7 +34,7 @@ public void start() {
   }
 
   public void sendResponse(Request newRequest, PrintWriter out) {
-    Controller controller = Router.dispatch(newRequest);
+    Controller controller = Router.dispatch(newRequest, publicDirectory);
```
aaronlahey: I'm surprised to see a change to the router because of a change to the way requests are handled.

src/main/java/me/karakelley/http/HttpServer.java
```diff
@@ -6,21 +6,23 @@
 import java.io.*;
 import java.net.ServerSocket;
 
-class HttpServer {
+public class HttpServer {
   private final int port;
-  private final Logger logger = LoggerFactory.getLogger("Server");
+  private final Logger logger = LoggerFactory.getLogger(HttpServer.class);
```
aaronlahey: All the loggers should really be `static`.

src/main/java/me/karakelley/http/HttpServer.java
```diff
@@ -6,21 +6,23 @@
...
+  private final PublicDirectory publicDirectory;
   private ServerSocket serverSocket;
 
-  public HttpServer(int port) {
+  public HttpServer(int port, PublicDirectory publicDirectory) {
```
aaronlahey: This doesn't seem very open-closed. Is there a way you can structure the code so that the server class doesn't need to change when we change how requests are handled?

src/main/java/me/karakelley/http/PublicDirectory.java
```diff
@@ -0,0 +1,34 @@
...
+
+  public String getPath() {
+    return path;
+  }
+
+  public List<String> getFiles() {
+    List<String> allFiles = new ArrayList<>();
+
+    File[] files = documentRoot.listFiles();
+    for (File file : files)
```
aaronlahey: https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#stream-T:A-

src/main/java/me/karakelley/http/ServerConfiguration.java
```diff
@@ -1,55 +1,66 @@
...
+
 public class ServerConfiguration {
-  private final String[] args;
+  private final Logger logger = LoggerFactory.getLogger(ServerConfiguration.class);
   private final int DEFAULT_PORT = 0;
-  private Integer port = -1;
-  final Logger logger = LoggerFactory.getLogger(ServerConfiguration.class);
+  private static final String DEFAULT_PATH = "nodirectory";
+  private final ArgumentParser parser;
+  private final Exit systemExit;
```
aaronlahey: It seems really strange to me that a class that's responsible for providing configuration to the server would have the authority to exit the system.

src/test/java/me/karakelley/http/ArgumentParserTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+class ArgumentParserTest {
+
+  private static Logger rootLogger;
+  private static InMemoryAppender inMemoryAppender;
+  private ArgumentParser parser;
+
+  @BeforeEach
+  void setUp() {
+    rootLogger = (Logger) LoggerFactory.getLogger("ROOT");
```
aaronlahey: Since you have lambdas you can really clean up this pattern of setting up resources, capturing some output, and then restoring resources. I'd love to see something like...

```
List<WhateverTypeTheLogMessageIs> = captureLogging(() -> {
  // run your code.
});
// assert against your captured log messages here...
```

src/test/java/me/karakelley/http/HttpServerTest.java
```diff
@@ -153,7 +154,7 @@ void testPortTooLarge() throws Exception {
 
   @Test
   void testSocketNotAvailable() {
-    httpServer = new HttpServer(4050);
+    httpServer = new HttpServer(4050, publicDirectoryMock);
```
aaronlahey: When you have to go back and change every test because you're adding new functionality, it's a good time to take a step back and ask yourself if its really necessary.

src/test/java/me/karakelley/http/Mocks/ArgumentParserMock.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public ArgumentParserMock(String[] args) {
+    super(args);
+    this.args = args;
+  }
+
+  public Map<String,String> getArgsHash() {
+    return argsHash;
+  }
+
```
aaronlahey: spacing

src/test/java/me/karakelley/http/Mocks/ArgumentParserMock.java
```diff
@@ -0,0 +1,65 @@
+package me.karakelley.http.Mocks;
+
+import me.karakelley.http.utility.ArgumentParser;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ArgumentParserMock extends ArgumentParser {
```
aaronlahey: Why do you need to mock the argument parser?

src/main/java/me/karakelley/http/ServerConfiguration.java
```diff
@@ -1,55 +1,66 @@
...
+  private Integer port;
+  private String directory = "";
+  Map<String, String> argsHash;
 
-  public ServerConfiguration(String[] args) {
-    this.args = args;
+  public ServerConfiguration(ArgumentParser parser, Exit systemExit) {
```
aaronlahey: Why does the server configuration have to know about command line arguments? I almost feel like the `ServerConfiguration` should just be a simple DTO which is built up from various sources.

I could also see an argument (no pun intended) for the `ServerConfiguration` being an interface and having one particular implementation of that being `CommandLineArguments implements ServerConfiguration` and then in tests you could just create simple implementations of `ServerConfiguration` like...


```
ServerConfiguration config = new ServerConfiguration() {
  @Override
  public int getPort() {
    return 0;
  }

  // ...
}
```

src/test/java/me/karakelley/http/Mocks/PublicDirectoryMock.java
```diff
@@ -0,0 +1,22 @@
...
+
+
+import me.karakelley.http.PublicDirectory;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PublicDirectoryMock extends PublicDirectory {
+
+
```
aaronlahey: extra space

src/test/java/me/karakelley/http/Mocks/PublicDirectoryMock.java
```diff
@@ -0,0 +1,22 @@
+package me.karakelley.http.Mocks;
+
+
+import me.karakelley.http.PublicDirectory;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PublicDirectoryMock extends PublicDirectory {
```
aaronlahey: Why do you need a mock public directory?

src/test/java/me/karakelley/http/PublicDirectoryTest.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class PublicDirectoryTest {
+
+  @Test
+  void listFiles() {
+    PublicDirectory directory = new PublicDirectory("./src/test/resources/directory/");
```
aaronlahey: Hard-coding paths in tests is a smell. You're probably dependent on your machine, or how the code is run (intellij vs compiled class files vs jars, etc).

It would be better to create a temp directory in your test, add some files to it, and then delete the temp directory once you're done. The suggestion above about the logging appender could also apply here...

```
// no directory exists
withTempDirectory(dir -> {
  // add a couple files
  // make your assertions
});
// no directory exists
```

aaronlahey (changes requested):

src/test/java/me/karakelley/http/ArgumentParserTest.java
parkerpersaud: I'm not quite sure what you mean - use a lambda in the `setup` or in each unit test?

PR #11, opened 2018-03-27
Feat navigable directories

parkerpersaud:
https://trello.com/c/x8Nqq64K/74-275-make-directories-navigable


src/main/java/me/karakelley/http/PublicDirectory.java
```diff
@@ -24,16 +24,24 @@ public static PublicDirectory create(String path) {
...
 
-  public static List<String> getFiles() {
-    File[] files = documentRoot.listFiles();
-    return Arrays.stream(files)
-            .filter(File::isFile)
-            .map(File::getName)
+  public static List<File> getDirectoriesAndFiles(String requestedResource) {
+    File directory = Paths.get(documentRoot + requestedResource).toFile();
+    return Arrays.stream(directory.listFiles())
+            .filter(file -> file.isDirectory() || file.isFile())
```
mikeknep: Looks like this could use `validResource`

src/main/java/me/karakelley/http/controllers/Router.java
```diff
@@ -16,6 +17,9 @@ public Router route(String path, Controller controller) {
 
   @Override
   public Response respond(Request request) {
+    if (PublicDirectory.validResource(request.getPath())) {
+      return mappings.getOrDefault(request.getPath(), new StaticFilesController()).respond(request);
```
mikeknep: The structure of this method seems a little strange to me, I think because you're checking which "fallback" controller to use when you may not need to use either fallback (if there is an explicit mapping for the request path). There's probably a way to make this read a little more descriptively in terms of "route priorities" (for lack of a better term). (Might require abandoning `getOrDefault`.)

mikeknep: Also, beyond just code clarity, it's probably good to avoid the I/O hit of checking whether something is a valid resource if you can.

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
```diff
@@ -4,27 +4,46 @@
...
 
   public Response respond(Request request) {
+    String requestedResource = request.getPath();
+
     Response response = new Response();
     if (request.validRequestLine() && availableActions.contains(request.getMethod())) {
-      return displayFilesResponse(response);
+      return serveDirectoryResponse(response, requestedResource);
     } else {
       return new InvalidRequestController().respond(request);
```
mikeknep, line 32: This might not be the best time to address this, but I don't think this `StaticFilesController` class should have a dependency on another controller—aren't they basically "peers"? It'd be nice if you could guarantee that if you enter this class, only one thing can happen (i.e. there is _only_ a "happy path").

If you want to hold on that and do it in a separate PR I'm OK with that.

src/test/java/me/karakelley/http/HttpServerTest.java
```diff
@@ -257,7 +274,8 @@ void testDisplaysFilesAtRoot() {
       } catch (IOException e) {
         e.printStackTrace();
       }
-      assertTrue(response.size() == 6);
+      assertTrue(response.contains("Content-Type: text/html"));
+      assertTrue(response.size() > 4);
```
mikeknep: What's this magic number `4` all about? And why the change from a precise `==` to a more lenient `>`?

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
```diff
@@ -4,27 +4,46 @@
...
-    response.setHeaders("Content-Type", "text/plain");
+    response.setBody(String.join("", getFilesAndDirectoryLinks(requestedResource)));
+    response.setHeaders("Content-Type", "text/html");
     response.setHeaders("Content-Length", String.valueOf(response.getBody().getBytes().length));
     return response;
   }
+
+  private ArrayList<String> getFilesAndDirectoryLinks(String resource) {
+    List<File> files = PublicDirectory.getDirectoriesAndFiles(resource);
+    ArrayList<String> directories = new ArrayList<>();
```
mikeknep: nit: this variable might be more accurately named `resources`, since you could have files or directories

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
```diff
@@ -4,27 +4,46 @@
...
+    for (File file : files) {
+      if (file.isDirectory()) {
+        directories.add(HtmlGenerator.createLink(file));
```
mikeknep: This feels pretty concrete/coupled to me

src/main/java/me/karakelley/http/utility/HtmlGenerator.java
```diff
@@ -0,0 +1,15 @@
+package me.karakelley.http.utility;
+
+import me.karakelley.http.PublicDirectory;
+
+import java.io.File;
+
+public class HtmlGenerator {
```
mikeknep: No tests for this?

src/main/java/me/karakelley/http/utility/HtmlGenerator.java
```diff
@@ -0,0 +1,15 @@
...
+  public static String createLink(File fileName) {
+    return "<p><a href=\"/" + PublicDirectory.relativePath(fileName) + "\">" + fileName.getName() + "</a></p>";
```
mikeknep: There is some tight coupling between `HtmlGenerator`, `PublicDirectory`, and `StaticFilesController`

src/main/java/me/karakelley/http/controllers/Router.java
parkerpersaud: Can you expand on this? Is there a way validate and avoid hitting I/O altogether? At some point you have to check the file system to verify it exists, right?

mikeknep: Ah, yeah I was unclear. If the path is in your predefined route mappings, you don't have to check the filesystem. For example, the `GET /redirectme` should probably do its redirection without having to look at the filesystem (at least not until it handles the redirected-to target, `/`).

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
parkerpersaud: Yeah definitely, I think if I can abstract the request validation I can remove that dependency. I am going to add it to my list.

src/main/java/me/karakelley/http/PublicDirectory.java
```diff
@@ -24,16 +24,32 @@ public static PublicDirectory create(String path) {
...
+    File directory = Paths.get(documentRoot + requestedResource).toFile();
+    if (validDirectory(requestedResource)) {
+      return Arrays.stream(directory.listFiles())
+              .filter(file -> file.isDirectory() || file.isFile())
+              .collect(Collectors.toList());
+    }
+    return null;
+  }
+
+  public boolean validResource(String path) {
```
aaronlahey: What does `valid` mean here?

src/main/java/me/karakelley/http/controllers/Router.java
```diff
@@ -16,6 +26,17 @@ public Router route(String path, Controller controller) {
...
+      return mappings.get(request.getPath()).respond(request);
+    } else {
+      checkRequestedPath(request);
+      return mappings.getOrDefault(request.getPath(), new InvalidRequestController()).respond(request);
+    }
+  }
+
+  private void checkRequestedPath(Request request) {
+    if (publicDirectory != null && publicDirectory.validResource(request.getPath())) {
+      mappings.put(request.getPath(), new StaticFilesController(publicDirectory, new HtmlGenerator()));
```
aaronlahey: Why does the router need to know anything about static files controller and html?

src/main/java/me/karakelley/http/PublicDirectory.java
```diff
@@ -24,16 +24,32 @@ public static PublicDirectory create(String path) {
...
-            .map(File::getName)
-            .collect(Collectors.toList());
+  public List<File> getDirectoriesAndFiles(String requestedResource) {
+    File directory = Paths.get(documentRoot + requestedResource).toFile();
+    if (validDirectory(requestedResource)) {
+      return Arrays.stream(directory.listFiles())
+              .filter(file -> file.isDirectory() || file.isFile())
+              .collect(Collectors.toList());
+    }
+    return null;
```
mikeknep: 1. This particular instance isn't glaringly problematic, but in general I don't like multiple return statements at "different levels". What I mean is: I don't want to have to look through the entire body of a method to find out where it might return early. There are some exceptions, but frankly this isn't one of them :) A simple `else` block at the very least would improve this, because then the last "clause" of your method is an `if/else` block. It's subtle, but you could imagine something like this:
```ruby
def do_some_stuff(x)
  y = x + 1
  if y.even?
    return false
  end
  #
  # ten more lines of code here
  #
  return true
end
```

2. I'd recommend avoiding `return null` as much as possible. `null` is a huge PITA, as is checking for it throughout a codebase. I'd think about whether this method could work returning either an empty `List<File>` or an `Option<List<File>>`

src/test/java/me/karakelley/http/HttpServerTest.java
```diff
@@ -257,7 +274,8 @@ void testDisplaysFilesAtRoot() {
       } catch (IOException e) {
         e.printStackTrace();
       }
-      assertTrue(response.size() == 6);
+      assertTrue(response.contains("Content-Type: text/html"));
+      assertTrue(response.size() == 5);
```
mikeknep: I still don't really know what this means. Why five?
1. "Magic numbers" in tests get confusing. I think an example might be helpful here—pardon my Ruby pseudo-code:
```ruby
it "lists all the files" do
  test_files = ["hello.txt", "world.txt", "bonjour.txt", "monde.txt"]
  test_files.each { |test_file| seed_test_file(test_file) }
  do_some_stuff
  
  expect(response.size).to eq(4)
  expect(response.size).to eq(test_files.length)
end
```
The second `expect` line is more self-documenting than the first: "The size of the response should be the same as the number of test files I made" vs. *looks around in test* "well, this array has 4 elements, that's probably what's going on."
2. In this particular case I'm genuinely not sure where five comes from. Have you seeded five files in the root directory for this test somehow? It looks like there are only two.

src/main/java/me/karakelley/http/PublicDirectory.java
```diff
@@ -24,16 +24,32 @@ public static PublicDirectory create(String path) {
...
+    return (directory.exists()) && directory.isDirectory() || directory.isFile();
+  }
+
+  public String relativePath(File file) {
+    return documentRoot.toURI().relativize(file.toURI()).getPath();
+  }
+
+  private boolean validDirectory(String path) {
+    File directory = Paths.get(documentRoot + path).toFile();
+    return directory.exists() && directory.isDirectory();
```
mikeknep: There still seems to be a lot of duplication between `validDirectory` and `resourceExists`, and that you'll have to change both methods together if you find a bug.

src/main/java/me/karakelley/http/controllers/Application.java
```diff
@@ -3,26 +3,36 @@
...
   public Application(PublicDirectory publicDirectory) {
-    router.route("/", new StaticFilesController());
+    this.publicDirectory = publicDirectory;
+    router.route("/", new StaticFilesController(publicDirectory, new HtmlGenerator()));
     setupCommonRoutes();
   }
 
   @Override
   public Response respond(Request request) {
+    checkRequestedPath(request);
```
aaronlahey: Check requested path for what? Why are we checking it?

src/main/java/me/karakelley/http/controllers/Application.java
```diff
@@ -3,26 +3,36 @@
...
     return router.respond(request);
   }
 
   private void setupCommonRoutes() {
     router.route("/redirectme", new RedirectController());
   }
+
+  private void checkRequestedPath(Request request) {
+    if (publicDirectory != null && publicDirectory.resourceExists(request.getPath())) {
+      router.route(request.getPath(), new StaticFilesController(publicDirectory, new HtmlGenerator()));
```
aaronlahey: I was not expecting to see routes dynamically added to the controller every time a request comes in.... especially in a method called `checkRequestedPath`. Why do you need to do this?

src/main/java/me/karakelley/http/controllers/Router.java
```diff
@@ -16,6 +18,10 @@ public Router route(String path, Controller controller) {
 
   @Override
   public Response respond(Request request) {
-    return mappings.getOrDefault(request.getPath(), new InvalidRequestController()).respond(request);
+    if (mappings.containsKey(request.getPath())) {
+      return mappings.get(request.getPath()).respond(request);
+    } else {
+      return mappings.getOrDefault(request.getPath(), new InvalidRequestController()).respond(request);
```
aaronlahey: Why `getOrDefault` in the `else` branch. Won't it always be the default?

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
```diff
@@ -4,27 +4,55 @@
...
+import java.util.List;
 
 public class StaticFilesController implements Controller {
-  private final ArrayList availableActions = new ArrayList<>(Collections.singleton("GET"));
+
+  public final ArrayList availableActions = new ArrayList<>(Collections.singleton("GET"));
+  private final HtmlGenerator htmlGenerator;
+  private final PublicDirectory publicDirectory;
+
+  public StaticFilesController(PublicDirectory publicDirectory, HtmlGenerator htmlGenerator) {
```
aaronlahey: Why do you need to inject an instance of `HtmlGenerator`?

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
```diff
@@ -4,27 +4,55 @@
...
       return new InvalidRequestController().respond(request);
     }
   }
 
-  private Response displayFilesResponse(Response response) {
+  private Response serveDirectoryResponse(Response response, String requestedResource) {
     response.setStatus(Status.OK);
-    response.setBody(String.join("\n", PublicDirectory.getFiles()));
-    response.setHeaders("Content-Type", "text/plain");
+    response.setBody(String.join("", getFilesAndDirectoryLinks(requestedResource)));
```
aaronlahey, line 38: Why join with empty string?

src/main/java/me/karakelley/http/controllers/Application.java
parkerpersaud: I need to check if the requested resource exists

src/main/java/me/karakelley/http/controllers/Application.java
parkerpersaud: I thought it made sense to create the routes as they come in - Initially, I created the function in `Router` to avoid hitting I/O every time thus I was creating dynamic routes as requests for valid resources came in. I realize now, after moving it to `Application` that it is no longer serving the purpose I intended. I currently see two scenarios, either add routes dynamically as they come in or check if the file exists every time no matter what.

src/main/java/me/karakelley/http/controllers/Router.java
parkerpersaud: It always defaults if there is a directory passed in when the server starts. But, If there is no directory and you try to hit a route that hasn't been predefined (i.e. `/` or `/redirectme` then it will default.

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
parkerpersaud: I wanted to generate html when I set the response body. Before, I was using a static method to generate the HTML but it was tightly coupled so I decided to use dependency injection.

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
parkerpersaud: It's an list of html elements -  I initially was joining on a newline but the browser renders the html and newlines at that point didn't seem necessary.

src/main/java/me/karakelley/http/controllers/Application.java
aaronlahey: Ok. Cool. I'd say a better function name would be `checkIfRequestedResourceExists`. That name encapsulates the "why" but it still doesn't really explain "ok, then what." You can generally say that something will return a `boolean` by phrasing it like a predicate `requestedResourceExists` and you can convey that a method will throw by prepending it with `ensure` or `assert`. `ensureTheRequestedResourceExists` or `assertTheRequestedResourceExists`. In this case I think you're just adding a route to the router, so I'd probably rename this altogether and go with `addRouteForRequestedResource`.

src/main/java/me/karakelley/http/controllers/Application.java
aaronlahey: Ahhhh. I see. I think the avoiding IO on every request is a good thing to think about. There are probably cleaner ways to go about it, though. Can you think of a way you could use the [Proxy Pattern](https://en.wikipedia.org/wiki/Proxy_pattern) to avoid the IO?

src/main/java/me/karakelley/http/controllers/Router.java
aaronlahey: I've thought about this for like 5 minutes now and I still don't understand why you can't replace lines 21-25 with just...

```
return mappings.getOrDefault(request.getPath(), new InvalidRequestController()).respond(request);
```

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
aaronlahey: Ohhhh. This is joining the elements of a list with an empty string. I thought you were just merging two `String`s...one of which was `""`. That makes much more sense.

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
aaronlahey: I would argue this is still pretty tightly coupled (although less so than using a `static` method). You could change this implementation if you sub-classed `HtmlGenerator` and overrode all its methods. That would be pretty weird, though, so it's probably better to depend on an `interface` or `abstract class` to reduce coupling.

But, lets pretend `HtmlGenerator` *is* an `interface`. That naming alone is causing some coupling. What if I wanted to return the directory contents as JSON? What about CSV? XML? Do you really need an `HtmlGenerator` or is there something even more abstract you could depend on?

src/main/java/me/karakelley/http/controllers/Router.java
parkerpersaud: Yeah,  you're right. For some reason I kept trying to remove line 24 instead of 21-23

src/main/java/me/karakelley/http/controllers/Application.java
parkerpersaud: After reading about it, I think I understand how the pattern works but what I don't understand is how using it avoids the IO. Eventually you're still hitting IO to check if the resource exists, right?

src/main/java/me/karakelley/http/controllers/Router.java
aaronlahey: 👍

src/main/java/me/karakelley/http/controllers/Application.java
aaronlahey: https://en.wikipedia.org/wiki/Memoization

aaronlahey: Does that spark any ideas?

src/test/java/me/karakelley/http/HttpServerTest.java
```diff
@@ -198,6 +201,7 @@ void testRedirectsWhenPathIsRedirectme() throws Exception {
     client.connectWithTry("127.0.0.1", httpServer);
 
     ArrayList<String> response = client.sendMessage("GET /redirectme HTTP/1.1\r\n");
+    System.out.println(response);
```
aaronlahey: 😱

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
```diff
@@ -1,30 +1,58 @@
...
+    response.setHeaders("Content-Type", "text/html");
     response.setHeaders("Content-Length", String.valueOf(response.getBody().getBytes().length));
     return response;
   }
+
+  private ArrayList<String> getFilesAndDirectoryLinks(String resource) {
+    List<File> files = publicDirectory.getDirectoriesAndFiles(resource);
+    ArrayList<String> resources = new ArrayList<>();
+    if (files != null) {
```
aaronlahey: Does `getDirectoriesAndFiles` ever return `null`? It looks like it just returns an empty list if the requested resource doesn't exist.

src/main/java/me/karakelley/http/FileSystem/FileFinderProxy.java
```diff
@@ -0,0 +1,22 @@
+package me.karakelley.http.FileSystem;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class FileFinderProxy implements FileFinder {
```
aaronlahey: I think adding `Proxy` to this name is probaby good/helfpul, but it's also really vague. What functionality does it provide? Does it provide authorization? Caching? Logging?

Note the [`HttpMessageConverter`](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/http/converter/HttpMessageConverter.html) interface from the Spring project. Notice how all the implementations describe what functionality they provide.

src/main/java/me/karakelley/http/controllers/Application.java
```diff
@@ -1,28 +1,38 @@
...
     return router.respond(request);
   }
 
   private void setupCommonRoutes() {
     router.route("/redirectme", new RedirectController());
   }
+
+  private void addRouteForRequestedResource(Request request) {
+    if (publicDirectory != null && publicDirectory.resourceExists(request.getPath())) {
+      router.route(request.getPath(), new StaticFilesController(publicDirectory, new HtmlGenerator()));
```
aaronlahey: I still think this is a really strange solution to this problem. Can you just call into the `StaticFilesController` and let it return the files which correspond to the path?

src/test/java/me/karakelley/http/utility/HtmlGeneratorTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.nio.file.Path;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class HtmlGeneratorTest {
+
+  @Test
+  void testCreateLink() {
```
aaronlahey: Needs a rename.

src/test/java/me/karakelley/http/utility/HtmlGeneratorTest.java
```diff
@@ -0,0 +1,32 @@
...
+    TempFilesHelper.withTempDirectory(directory -> {
+      Path pathOne = TempFilesHelper.createTempFile(directory);
+
+      assertTrue(htmlGenerator.displayDirectories(pathOne.toFile(), PublicDirectory.create("/", new FileFinderProxy(new RealFileFinder()))).contains("<p><a href="));
+      assertTrue(htmlGenerator.displayDirectories(pathOne.toFile(), PublicDirectory.create("/", new FileFinderProxy(new RealFileFinder()))).contains("</a></p>"));
+    });
+  }
+
+  @Test
+  void testCreateParagraph() {
```
aaronlahey: Needs a rename.

src/test/java/me/karakelley/http/utility/HtmlGeneratorTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+class HtmlGeneratorTest {
+
+  @Test
+  void testCreateLink() {
+    HtmlGenerator htmlGenerator = new HtmlGenerator();
+    TempFilesHelper.withTempDirectory(directory -> {
+      Path pathOne = TempFilesHelper.createTempFile(directory);
+
+      assertTrue(htmlGenerator.displayDirectories(pathOne.toFile(), PublicDirectory.create("/", new FileFinderProxy(new RealFileFinder()))).contains("<p><a href="));
```
aaronlahey: For a test like this I'd probably prefer to find a way to assert *exactly* what I wanted to see as the return value.

```
assertEquals("<p><a href="/some/path/here">file-name.txt</a></p>")
```

I feel like every time you make a test dynamic like this you increase the risk of making the same bad assumption in both the assertion and the production code, which leads to bugs. It's not always bad, though. You just need to be aware of those risks.

I expect it'll be hard to test because this depends on an actual file being on the filesystem. At this level, it might be better to make it more of a "pure function" (something that takes a value and returns a value) rather than something that depends on global state (the fileystem in this case).

aaronlahey (commented):
Nearing the end of my suggestions here, don't worry!

src/main/java/me/karakelley/http/controllers/Application.java
parkerpersaud: I don't quite understand what you are suggesting is strange about the solution. Validating the resource exists and if there's a directory or how I'm adding the route?

src/test/java/me/karakelley/http/utility/HtmlGeneratorTest.java
parkerpersaud: I did fix this in the next PR for opening text files. I altered the `TempFilesHelper` to give me more concrete names so I can make better assertions (just like you suggested).

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
```diff
@@ -1,30 +1,58 @@
...
+    response.setBody(String.join("", getFilesAndDirectoryLinks(requestedResource)));
+    response.setHeaders("Content-Type", "text/html");
     response.setHeaders("Content-Length", String.valueOf(response.getBody().getBytes().length));
     return response;
   }
+
+  private ArrayList<String> getFilesAndDirectoryLinks(String resource) {
+    List<File> files = publicDirectory.getDirectoriesAndFiles(resource);
+    ArrayList<String> resources = new ArrayList<>();
+    if (!files.isEmpty()) {
```
mikeknep: Previously you had to do a null check here. The nice thing about changing `PublicDirectory#getDirectoriesAndFiles` to always return a `List<Files>` (and having that list potentially be empty) is that the client doesn't need to make this check at all any more—iterating over an empty list is simply a no-op.
```
List<File> files = new ArrayList<>();
for (File file : files ) {
  println("Hello")
}
```
Nothing gets printed.

PR #12, opened 2018-03-29
Feat open files

parkerpersaud:
https://trello.com/c/fN2YvGMN/75-375-open-text-files


src/main/java/me/karakelley/http/ContentGeneration/ContentGenerator.java
```diff
@@ -6,5 +6,4 @@
 
 public interface ContentGenerator {
   String displayDirectories(File fileName, PublicDirectory publicDirectory);
```
mikeknep: As discussed in Slack, this name no longer makes sense.

src/main/java/me/karakelley/http/FileSystem/PublicDirectory.java
```diff
@@ -43,11 +46,32 @@ private PublicDirectory(String path, FileFinder fileFinderProxy) {
...
+
+  public boolean isFile(String requestedResource) {
+    return getPath(requestedResource).toFile().isFile();
+  }
+
+  public byte[] getFileContents(String requestedResource) {
+    Path file = getPath(requestedResource);
+    try {
+      return Files.readAllBytes(file);
+    } catch (IOException e) {
```
mikeknep, line 69: Under what circumstances would an `IOException` be thrown?

src/main/java/me/karakelley/http/HttpServer.java
```diff
@@ -37,19 +37,23 @@ public void start() {
...
     }
   }
 
-  public void sendResponse(Request newRequest, PrintWriter out) {
+  private void sendResponse(Request newRequest, OutputStream out) {
     Response response = controller.respond(newRequest);
-    out.write(response.deliver());
+    try {
+      out.write(response.deliver());
+    } catch (IOException e) {
```
mikeknep: Same question: describe what would have to happen for an `IOException` to get thrown on the line above.

src/main/java/me/karakelley/http/Response.java
```diff
@@ -17,17 +22,48 @@ public void setHeaders(String header, String value) {
...
+  }
+
+  public byte[] getBody() {
     if (body != null) {
       return body;
-    } else return "";
+    } else return "".getBytes();
+  }
+
+  public byte[] deliver() {
```
mikeknep: Does this method actually send the contents out? The name `deliver` suggests it does, but it does return a `byte[]` which makes me think that something else is probably actually doing the delivery (otherwise this could presumably be `void`). Also the body of this method seems like it's just doing some conversion.

src/main/java/me/karakelley/http/ResponseFormatter.java
```diff
@@ -1,31 +0,0 @@
-package me.karakelley.http;
-
-import java.util.Map;
-import java.util.stream.Collectors;
-
-public class ResponseFormatter {
```
mikeknep, line 6: Why did you decide to move all of this into `Response`?

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
```diff
@@ -27,30 +26,45 @@ public Response respond(Request request) {
...
+    response.setHeaders("Content-Length", String.valueOf(response.getBody().length));
     return response;
   }
 
   private ArrayList<String> getFilesAndDirectoryLinks(String resource) {
     List<File> files = publicDirectory.getDirectoriesAndFiles(resource);
     ArrayList<String> resources = new ArrayList<>();
     for (File file : files) {
-      if (file.isDirectory()) {
+      if (file.isDirectory() || file.isFile()) {
```
mikeknep: Is anything inside the `files` list _not_ a directory or file? Do you need this check at all?

src/test/java/me/karakelley/http/HttpServerTest.java
```diff
@@ -277,7 +277,37 @@ void testDisplaysFilesAtRoot() {
...
+    });
+  }
+
+  @Test
+  void testReturnsFileContentsRequested() {
+    TempFilesHelper.withTempDirectory(directory -> {
+      Path fileOne = TempFilesHelper.createTempFile(directory, "/test1");
+      writeToFile("Hello World", fileOne);
+      ServerConfiguration config = new ServerConfiguration();
+      config.setController(new Application(PublicDirectory.create(directory.toString(), new FileFinderCache(new RealFileFinder()))));
```
mikeknep: I'm a little confused about why the argument to `setController` is an `Application`...

src/test/java/me/karakelley/http/HttpServerTest.java
```diff
@@ -277,7 +277,37 @@ void testDisplaysFilesAtRoot() {
...
+      try {
+        client.connectWithTry("127.0.0.1", httpServer);
+      } catch (InterruptedException | IOException e) {
+        e.printStackTrace();
+      }
+
+      ArrayList<String> response = null;
+      try {
+        response = client.sendMessage("GET /test1.txt HTTP/1.1\r\n");
+      } catch (IOException e) {
```
mikeknep, line 306: Does this ever get thrown and caught when you run your tests?

src/test/java/me/karakelley/http/controllers/StaticFilesControllerTest.java
```diff
@@ -22,39 +21,65 @@
...
+      assertEquals("Hello World", new String(response.getBody()));
+    });
   }
 
   private LineReader newBufferedReader(String request) {
     ByteArrayInputStream inputStream = new ByteArrayInputStream(request.getBytes());
     return new BufferedLineReader(new InputStreamReader(inputStream));
   }
+
+  private void writeToFile(String text, Path file) {
```
mikeknep, line 75: This is in two different test files, and probably belongs in neither. There are a couple different styles to choose from that would be a little more ergonomic. One suggestion is something like:
```
TempFilesHelper.createTempFile(directory, "/test1").withContents("Hello World")
```

src/test/java/me/karakelley/http/HttpServerTest.java
parkerpersaud: Not that I've noticed - I kept getting compiler errors without the `try/catch` block within the lambda and adding it to the method signature wouldn't work either.

src/test/java/me/karakelley/http/HttpServerTest.java
parkerpersaud: It doesn't have to be `Application` but since these are my end-to-end tests and `Application` is what I initialize my program with it made sense in order to test it end-to-end.

src/main/java/me/karakelley/http/ResponseFormatter.java
parkerpersaud: When I was changing the functionality of `ResponseFormatter` it made sense that this belonged in `Response`.

src/main/java/me/karakelley/http/controllers/StaticFilesController.java
parkerpersaud: That's a good point - At this point everything in the list is already a file or directory so I don't need to check it again

src/test/java/me/karakelley/http/HttpServerTest.java
mikeknep: Ahh, I didn't see that `Application` implements the `Controller` interface. Let's talk about this outside the context of this PR.

src/test/java/me/karakelley/http/HttpServerTest.java
mikeknep: Fair enough, this one might be unavoidable. Still curious about some of the other `try/catch` blocks I commented on.

src/main/java/me/karakelley/http/HttpServer.java
parkerpersaud: I believe in this case, it would be thrown if the output stream has been closed.

PR #13, opened 2018-03-29
Fix null pointer exception

parkerpersaud:
Fix logging of null pointer exception

### DIFF
--- a/src/main/java/me/karakelley/http/Request.java
+++ b/src/main/java/me/karakelley/http/Request.java
@@ -82,10 +82,11 @@ public class Request {
 
   private void setRequestLine() {
     String requestLine;
-    String[] parsedLine;
+    String[] parsedLine = new String[2];
     try {
-      requestLine = reader.readLine();
-      parsedLine = requestLine.replaceAll("\\s+", " ").split(" ");
+      if ((requestLine = reader.readLine()) != null) {
+        parsedLine = requestLine.replaceAll("\\s+", " ").split(" ");
+      }
       if (parsedLine.length == 3) {
         requestMethod = parsedLine[0];
         requestPath = parsedLine[1];
@@ -95,5 +96,4 @@ public class Request {
       throw new RuntimeException(e);
     }
   }
-
 }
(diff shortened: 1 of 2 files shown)


src/main/java/me/karakelley/http/Request.java
```diff
@@ -82,10 +82,11 @@ private void setHeaders() {
 
   private void setRequestLine() {
     String requestLine;
-    String[] parsedLine;
+    String[] parsedLine = new String[2];
```
mikeknep, line 85: Can you explain why this change is necessary?

src/main/java/me/karakelley/http/Request.java
```diff
@@ -82,10 +82,11 @@ private void setHeaders() {
...
     try {
-      requestLine = reader.readLine();
-      parsedLine = requestLine.replaceAll("\\s+", " ").split(" ");
+      if ((requestLine = reader.readLine()) != null) {
+        parsedLine = requestLine.replaceAll("\\s+", " ").split(" ");
```
mikeknep, line 88: Just noticed this—it's ancillary to this PR since this line is just an indentation change, but: is there a reason you're not just splitting on `\\s+`? The sub from that to `" "` seems superfluous.

src/test/java/me/karakelley/http/RequestTest.java
```diff
@@ -97,6 +96,17 @@ void something() throws Exception {
     }
   }
 
+  @Test
+  void testNullPointerExceptionIsNotThrown() {
```
mikeknep, line 100: I'd rather see this test define what _does_ happen in this situation, rather than just calling out one thing that doesn't happen, for a few reasons:
1. This high-level situation (a socket connection opening up but no data coming through) seems plausible (in fact, we've seen it happen when poking around the server already), and so this test could better serve as documentation of expected behavior. As-is, I don't know what should happen; I only know one thing that won't.
2. There are other app-crashing things that can happen beyond exceptions (see Error and Throwable), so I feel like this test doesn't sufficiently prove that your server won't blow up when this happens.

src/main/java/me/karakelley/http/Request.java
parkerpersaud: Because of the first `if` block where it checks for null, since we may not hit that line and initialize `parsedLine` then it may not be initialized therefore it's initialized on line 85, outside of the `if` block.

PR #14, opened 2018-04-02
Refactor requests and responses

parkerpersaud:
- Refactored the overall request validation to happen at a higher level instead of in the controllers
- Changed controllers to handlers 
- Created new Exceptions package 

There is no new functionality. The only small thing that changed is that a 400 Bad Request is returned for syntactically invalid requests. 

Some other things that I'd still like to refactor include relying more on the type system and refactoring the packages


src/main/java/me/karakelley/http/Request.java
```diff
@@ -1,23 +1,20 @@
...
   private final int port;
   private String requestMethod;
   private String requestPath;
   private String requestProtocol;
-  final Map<String, String> headers = new HashMap<>();
+  private Map<String, String> headers;
 
-  public Request(LineReader reader, int port) {
-    this.reader = reader;
+  public Request(String requestMethod, String requestPath, String requestProtocol, Map<String, String> headers, int port) {
```
mikeknep, line 12: I like how simple this class has become—it is now an immutable representation of data coming into the server. Really nice. Small q: do you need the port for anything? It seems a little unrelated to the rest of the data (per request vs. per server)

src/main/java/me/karakelley/http/RequestBuilder.java
```diff
@@ -0,0 +1,10 @@
+package me.karakelley.http;
+
+import java.util.Map;
+
+public class RequestBuilder {
+
+  public static Request build(String requestMethod, String requestPath, String requestProtocol, Map<String, String> headers, int port) {
```
mikeknep, line 7: I like builders, but I'm curious: what made you reach for that concept here? The signatures are identical, so I'm wondering what advantage this provides over the caller of this function just knowing about `Request` directly.

src/main/java/me/karakelley/http/RequestParser.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+
+  private void parseRequestLine(LineReader reader) {
+    String requestLine;
+    String[] parsedLine;
+    try {
+      requestLine = reader.readLine();
+      parsedLine = requestLine.replaceAll("\\s+", " ").split(" ");
+      if (parsedLine.length == 3) {
```
mikeknep, line 52: If `parsedLine.length` is not 3, _and_ you don't encounter any exceptions, what happens?

src/main/java/me/karakelley/http/RequestValidator.java
```diff
@@ -0,0 +1,27 @@
+package me.karakelley.http;
+
+import java.util.Arrays;
+
+public class RequestValidator {
+  private String[] methods = new String[]{"GET", "POST"};
+
+  public boolean isValid(Request request) {
```
mikeknep, line 8: I really like that you separated this out into its own class. I wonder if it'd be worth adjusting it a bit so that `RequestValidator` operates on some temporary, pending request object—the idea being that your system cannot even create an invalid `Request` object at all. Maybe something with the Builder? I'm genuinely not sure, and maybe now is not the time to go overboard with it. But something to consider.

src/main/java/me/karakelley/http/ResponseFormatter.java
```diff
@@ -0,0 +1,46 @@
+package me.karakelley.http;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class ResponseFormatter {
```
mikeknep, line 8: This class is really nice and clean.

src/main/java/me/karakelley/http/exceptions/PublicDirectoryMissingException.java
```diff
@@ -0,0 +1,7 @@
+package me.karakelley.http.exceptions;
+
+public class PublicDirectoryMissingException extends RuntimeException{
```
mikeknep, line 3: What do you get by extending `RuntimeException` instead of just `Exception`? Is there a reason some extend one while others extend the other?

src/main/java/me/karakelley/http/handlers/RedirectHandler.java
```diff
@@ -0,0 +1,18 @@
...
+
+public class RedirectHandler implements Handler {
+
+  public Response respond(Request request) {
+    Response response = new MovedPermanently();
+    return setResponse(request, response);
+  }
+
+  private Response setResponse(Request request, Response response) {
+    response.setHeaders("Location", "http://" + request.getHostAndPort() + "/");
```
mikeknep, line 15: I _love_ how this is now only setting the header relevant to redirection and leaving the other, common headers to something else.

src/main/java/me/karakelley/http/handlers/Router.java
```diff
@@ -0,0 +1,30 @@
+package me.karakelley.http.handlers;
+
+import me.karakelley.http.Request;
+import me.karakelley.http.Response;
+
+import java.util.*;
+
+public class Router implements Handler {
+  private final Map<String, HashMap<List<String>, Handler>> mappings = new HashMap<>();
```
mikeknep: This is an interesting data structure—I'll be interested in seeing how it works out. Kinda hard to tell prior to handling any non-GET requests.

src/main/java/me/karakelley/http/responses/BadRequest.java
```diff
@@ -0,0 +1,10 @@
+package me.karakelley.http.responses;
+
+import me.karakelley.http.Response;
+
+public class BadRequest extends Response {
```
mikeknep, line 5: You'll find a lot of wariness around inheritance at 8th Light, for good reason. I like what you're going for here, but I think there are some alternatives to the inheritance approach. Tangentially related: check out [enums](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html)—might be a nice way to model your response codes.

src/main/java/me/karakelley/http/server/ConnectionHandler.java
```diff
@@ -0,0 +1,41 @@
...
+import me.karakelley.http.utility.LineReader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.DataOutputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.net.Socket;
+
+public class ConnectionHandler {
```
mikeknep, line 19: I think this was another great decision pulling this out from `HttpServer`—this class now defines the _what_, and `HttpServer` handles the _when_ (i.e. asynchronously upon receiving a request).

mikeknep (commented):
Haven't looked at the tests yet—will try to get some comments on those asap. Nice work!!

src/main/java/me/karakelley/http/Request.java
```diff
@@ -1,23 +1,20 @@
...
     this.port = port;
-    setRequestLine();
-    setHeaders();
+    this.headers = headers;
+    this.requestProtocol = requestProtocol;
```
aaronlahey: nitpick: I don't these variable names need to be prefixed with `request` since they're already properties of a `Request` object. Your getters seem to be fine, though.

src/main/java/me/karakelley/http/ResponseFormatter.java
```diff
@@ -0,0 +1,46 @@
...
+
+      out.close();
+      return out.toByteArray();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private byte[] formatHeaders(Map<String, String> headers) {
+    if (headers.isEmpty()) {
```
aaronlahey: Is this conditional necessary? I'd expect mapping over an empty collection then joining to already give you `""`.

src/main/java/me/karakelley/http/handlers/Application.java
```diff
@@ -1,34 +1,34 @@
...
+import me.karakelley.http.contentpresenter.HtmlPresenter;
 
-public class Application implements Controller {
+public class Application implements Handler {
   private final Router router = new Router();
   private PublicDirectory publicDirectory = null;
 
   public Application() {
-    router.route("/", new HelloWorldController());
+    router.route("/", new HelloWorldHandler(), new String[]{"GET"});
```
aaronlahey: I'm happy to see the request method become a part of the router.

A couple questions:

1. Why does a route take an array of methods?
2. Would it be possible to change the order of these arguments so it reads...

```
router.route("GET", "/", new HelloWorldHandler());
```

... I think that would make the interface read more like how one would say it in a conversation.

src/main/java/me/karakelley/http/handlers/Application.java
```diff
@@ -1,34 +1,34 @@
...
     setupCommonRoutes();
   }
 
   public Application(PublicDirectory publicDirectory) {
     this.publicDirectory = publicDirectory;
-    router.route("/", new StaticFilesController(publicDirectory, new HtmlGenerator()));
+    router.route("/", new StaticFilesHandler(publicDirectory, new HtmlPresenter()), new String[]{"GET"});
```
aaronlahey: Why have this route *and* line 26?

src/main/java/me/karakelley/http/handlers/Application.java
```diff
@@ -1,34 +1,34 @@
...
     setupCommonRoutes();
   }
 
   @Override
   public Response respond(Request request) {
     if (theRequestedResourceExists(request)) {
-      return new StaticFilesController(publicDirectory, new HtmlGenerator()).respond(request);
+      return new StaticFilesHandler(publicDirectory, new HtmlPresenter()).respond(request);
```
aaronlahey, line 26: Small note on behavior here:

If I request `/redirectme` *and* there is a static file called `redirectme` it should redirect and *not* serve the file. I'm ok with creating a card in Trello to change that behavior if that's not what is happening rather than fixing here.

src/main/java/me/karakelley/http/handlers/HelloWorldHandler.java
```diff
@@ -0,0 +1,18 @@
...
+import me.karakelley.http.Request;
+import me.karakelley.http.Response;
+
+public class HelloWorldHandler implements Handler {
+  private final String DEFAULT_RESPONSE = "Hello World";
+
+  @Override
+  public Response respond(Request request) {
+      Response response = new Ok();
+      response.setHeaders("Content-Type", "text/plain");
```
aaronlahey: Can all of these magic strings be refactored into constants?

src/main/java/me/karakelley/http/handlers/HelloWorldHandler.java
```diff
@@ -0,0 +1,18 @@
...
+      response.setHeaders("Content-Length", String.valueOf(DEFAULT_RESPONSE.getBytes().length));
```
aaronlahey: I don't think the `Handler`s should be responsible for setting the `Content-Length` header. It's making your `Handler`s a little *too* responsible for the correctness of the HTTP request that is ultimately sent to the client. For example, you could send a body with a `Content-Length` header that is either missing or wrong.

src/main/java/me/karakelley/http/handlers/Router.java
```diff
@@ -0,0 +1,30 @@
...
+
+  public Router route(String path, Handler handler, String[] methods) {
+    mappings.put(path, new HashMap<List<String>, Handler>() {{
```
aaronlahey: I'd be careful about doing this pattern in production code. Every time you do this it creates a new anonymous class (and an extra `*.class` file). I'll still do this in tests, though, if it really helps the readability. You don't need to change it, just be aware of it.

src/main/java/me/karakelley/http/responses/BadRequest.java
aaronlahey: My gut feel when I saw this pattern was that it would probably be kind of wild to have a subclass for [every possible status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status). That being said, it could enable a nice API for certain responses. For example...

```java
class MovedPermanently extends Response {
  // Can't even create a redirect response without a location
  public MovedPermanently(URI location) {
    super();
    setHeader("Location", location);
  }
}

class NoContent extends Request {
  // You can never have a body when you said you returned No Content.
  @Override
  public void setBody(byte[] body) {
    // Nothing...
  }
}
```

I'm honestly not sure if there are enough interesting invariants to enforce to make this pattern worth it. 🤷‍♂️ I'd be cool with moving to a single `Response` w/ an `enum`, but I'd also be cool with seeing where this pattern takes you (as long as you test at a high-enough level so you can easily bail out if needed).

src/test/java/me/karakelley/http/RequestValidatorTest.java
```diff
@@ -0,0 +1,38 @@
...
+  @Test
+  void inValidHTTPProtocol() {
+    Request request = new Request("GET", "/", "HTTP", new HashMap<>(), 0);
+    RequestValidator requestValidator = new RequestValidator();
+    assertFalse(requestValidator.isValid(request));
+  }
+
+  @Test
+  void inValidRequestMethod() {
+    Request request = new Request("GIT", "/", "HTTP/1.1", new HashMap<>(), 0);
```
aaronlahey, line 20: 💯

aaronlahey (approved):
This is a big improvement. Very well done! I'll approve with the feedback addressed.

src/main/java/me/karakelley/http/responses/BadRequest.java
parkerpersaud: I was hesitant to use inheritance but the reason I chose it was because it allowed me to dispatch a particular response type without having to worry about setting a particular status, etc. For instance, in `ConnectionHandler`, I can hand off a response object to `ResponseWriter` and let it do its thing.

```java 
 private void handleRequest(LineReader reader, OutputStream out, Handler handler, RequestParser requestParser, Socket clientSocket) {
    Response response;
    try {
      Request request = requestParser.parse(reader).buildRequest(clientSocket.getLocalPort());
      response = handler.respond(request);
      new ResponseWriter(out, response).deliver();
    } catch (InvalidRequestException e) {
      new ResponseWriter(out, new BadRequest()).deliver();
    }
  }
```

 I wanted to reach for enums but had a hard time figuring out how to use enums and keep my validations at this high level without worrying about implementation details.

src/main/java/me/karakelley/http/handlers/Application.java
parkerpersaud: If there is a directory called `/redirectme` and you request `localhost:5000/redirectme` you will be redirected. If, on the other hand, you request `localhost:5000/redirectme/` it will serve the contents of the directory.

src/main/java/me/karakelley/http/Request.java
parkerpersaud: I needed the port information for setting the `location` header for redirect responses. If I remember correctly, I needed it to support the redirect response for netcat

src/main/java/me/karakelley/http/RequestBuilder.java
parkerpersaud: Initially, I thought it would help to isolate the instantiation of `Request`. If for some reason, `Request` has to change, then the `RequestParser` would have to change for more than one reason. But as I type this, I realized I overlooked the fact that the argument list is identical so I'm not quite sure how much isolation there is at this point.

src/main/java/me/karakelley/http/RequestParser.java
parkerpersaud: The `requestMethod`, `requestPath` and `requestProtocol` remain null and soon after an `InvalidRequestException` is thrown. It could be thrown here now that I think of it. I will note that most of this parsing functionality has been changed in the `post-request` story but it still might be a good time to throw the exception at this point.

src/main/java/me/karakelley/http/exceptions/PublicDirectoryMissingException.java
parkerpersaud: If I'm honest, I didn't have a good reason but TIL `Exception` is checked while `RuntimeException` is unchecked. At this point, I don't want/need these particular exceptions to be checked so I changed them all to extend `RuntimeException`

src/main/java/me/karakelley/http/Request.java
mikeknep: Ah, makes sense. I was thinking that the host and port was somewhere else and redirect handler was just providing a relative path, but I think your approach may be more coherent if you ever wanted to redirect to a totally different host (for whatever reason).

src/main/java/me/karakelley/http/RequestBuilder.java
mikeknep: Right. Currently if request did change, this builder would change in exactly the same way. In fact, you'll probably run into this when you start handling POST requests, so let's leave it as-is in this PR and you can decide if/how you want it to change when you get to that point.

src/main/java/me/karakelley/http/RequestParser.java
mikeknep: I'm willing to leave it alone here and see how you change it in as a result of POST functionality

src/main/java/me/karakelley/http/exceptions/PublicDirectoryMissingException.java
mikeknep: So is it true that by making them unchecked, clients can call methods that potentially throw these exceptions without needing a `try/catch` block? In which case when one is thrown, it'll bubble up to something higher... how far would that go in your server today?

I've been fighting (and losing) a battle over exceptions in Scala with some client colleagues. There are some patterns I've wanted to try in that language to avoid exceptions as much as possible, but I recognize some downsides... And of course this is Java, which is different...

I dunno. @aaronlahey do you have any thoughts on this?

mikeknep (approved):

src/main/java/me/karakelley/http/Request.java
aaronlahey: It seems strange to have the port as an attribute on the request since the port is a property of some low-level protocol (TCP) and the `Request` is a part of some high-level protocol (HTTP), but I don't think it is really that strange in practice: https://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getLocalPort().

src/main/java/me/karakelley/http/exceptions/PublicDirectoryMissingException.java
aaronlahey: I have a lot of thoughts on this but no nice clean rule of thumb...

Having a bunch of named exceptions was a pattern I used pretty successfully at a client. It allows for a "poor man's pattern matching."

```
try {
  ...
} catch (FooWentWrong | BarWentWrong) {
  ...
} catch (BazWentWrong) {
  ...
} catch (QuuxWent Wrong) {
  ...
}
```

I think exceptions tend to make things more complicated than people realize...

1. They circumvent the control flow
2. You don't really know what's going to be thrown and for what reasons (unless they're checked exceptions but then you're polluting everything with `throws Exception`).
3. You have to be extra careful when you have resources you need to free (by using `try`/`finally` or `try(...)`.

But, they *can* be less complicated than result types in a language like java and they can be cheaper if you're doing really low-level things (you can avoid allocating an object if you just need to return an `int` or something).

PR #15, opened 2018-04-04
Feat post

parkerpersaud:
https://trello.com/c/8YxcjdKP/77-875-post-request


src/main/java/me/karakelley/http/RequestBuilder.java
```diff
@@ -4,7 +4,13 @@
 
 public class RequestBuilder {
 
-  public static Request build(String requestMethod, String requestPath, String requestProtocol, Map<String, String> headers, int port) {
-    return new Request(requestMethod, requestPath, requestProtocol, headers, port);
+  public static Request build(Map<String, String> requestLineAndBody, Map<String, String> headers, int port) {
```
aaronlahey: I think it would be cleaner, less brittle, and more type safe if you actually implemented the standard "builder pattern" and then `RequestParser` used that instead of `Map<String, String>`.

src/main/java/me/karakelley/http/RequestBuilder.java
```diff
@@ -4,7 +4,13 @@
...
+    return new Request(requestLineAndBody.get("method"),
```
mikeknep: Agreed with @aaronlahey's comment directly above this. Notice how, within the scope of this class, `"method", "path", "protocol", "body"` are all "magic strings", vs. private constants in `RequestParser`. A private val in a class suggests that it can be changed without affecting anything outside that class, and yet in this case, if you changed `RequestParser.METHOD` to `"Method"`, this would break.

src/main/java/me/karakelley/http/filesystem/PublicDirectory.java
```diff
@@ -71,7 +70,19 @@ public boolean isFile(String requestedResource) {
     }
   }
 
-  private Path getPath(String requestedResource) {
+  public void createFile(String path, String contents) throws IOException {
+    File newFile = getPath(path).toFile();
+    if (newFile.exists() && newFile.isFile() || newFile.isDirectory()) throw new IOException();
```
mikeknep: Minor style suggestion: any time I'm dealing with a conditional that has both "and"s and "or"s, I like to add explicit parentheses, even if not strictly necessary (based on the compiler's prioritization), just to make it clear to readers how the logic works. For example, in this line, it's not immediately obvious between these two:
```
if ((newFile.exists() && newFile.isFile()) || newFile.isDirectory())

if (newFile.exists() && (newFile.isFile() || newFile.isDirectory()))
```

src/main/java/me/karakelley/http/handlers/InvalidRequestHandler.java
```diff
@@ -2,12 +2,16 @@
 
 import me.karakelley.http.Request;
 import me.karakelley.http.Response;
-import me.karakelley.http.responses.NotFound;
 
 public class InvalidRequestHandler implements Handler {
 
+  private final Response response;
+
+  public InvalidRequestHandler(Response response) {
```
mikeknep, line 10: It seems a little unusual for a `...RequestHandler` to have a constructor that takes a `Response`. Isn't that what these kinds of objects are designed to create? How would one already exist?

src/main/java/me/karakelley/http/Main.java
```diff
@@ -29,7 +30,7 @@ public static void main(String[] args, BiConsumer<Map<String, String>, ServerCon
 
     configurationBiConsumer.accept(argsHash, serverConfiguration);
 
-    HttpServer server = new HttpServer(serverConfiguration, new RequestParser(new RequestValidator()), new ConnectionHandler());
+    HttpServer server = new HttpServer(serverConfiguration, new RequestReader(), new RequestParser(new RequestValidator()), new ConnectionHandler());
```
mikeknep: What made you decide to add the `RequestReader` as its own argument vs. "nesting" it with the `RequestParser` and `RequestValidator` somehow? I'm not sure if I have a preference; it just happened to stand out to me and I'm curious about what guided this decision.

src/main/java/me/karakelley/http/handlers/Application.java
```diff
@@ -16,22 +16,18 @@ public Application() {
 
   public Application(PublicDirectory publicDirectory) {
     this.publicDirectory = publicDirectory;
-    router.route("GET", "/",new StaticFilesHandler(publicDirectory, new HtmlPresenter()));
+    router.route("GET", "/", new StaticFilesHandler(publicDirectory, new HtmlPresenter()));
+    router.route("POST", new NewResourceHandler(publicDirectory));
+    router.route("GET", new StaticFilesHandler(publicDirectory, new HtmlPresenter()));
```
mikeknep: Why do you need this line and, two lines above, nearly the same line except with an additional `"/"` argument?

src/main/java/me/karakelley/http/responses/HttpMethod.java
```diff
@@ -0,0 +1,21 @@
+package me.karakelley.http.responses;
+
+import me.karakelley.http.exceptions.InvalidRequestException;
+
+public enum HttpMethod {
```
mikeknep: Why is this in the `responses` package?

src/main/java/me/karakelley/http/handlers/Router.java
```diff
@@ -32,9 +32,9 @@ private Handler getController(Request request) {
     HashMap<String, Handler> resource = staticMappings.get(request.getPath());
 
     if (resource == null) {
-      return dynamicMappings.getOrDefault(request.getMethod(), new InvalidRequestHandler(new NotFound()));
+      return dynamicMappings.getOrDefault(request.getMethodAsString(), new InvalidRequestHandler(new NotFound()));
```
mikeknep: Why not use the enum, since you have it now?

src/main/java/me/karakelley/http/Main.java
```diff
@@ -30,7 +27,7 @@ public static void main(String[] args, BiConsumer<Map<String, String>, ServerCon
 
     configurationBiConsumer.accept(argsHash, serverConfiguration);
 
-    HttpServer server = new HttpServer(serverConfiguration, new ParserFactory(), new ConnectionHandler());
+    HttpServer server = new HttpServer(serverConfiguration, new ConnectionHandler(), new RequestReaderFactory());
```
mikeknep: Just curious: why did you choose to use a factory here?

parkerpersaud: I need a new instance of `HttpRequestReader` for each new connection and didn't want to go stateless with this class so I thought instead of instantiating it inside of `ConnectionHandler` I could pass something in that was responsible for creating a new `HttpRequestReader` on each connection.

PR #16, opened 2018-04-09
Bug dot paths

parkerpersaud:
https://trello.com/c/2zKgPQhh/97-375-fix-requests-with-dots-404

### DIFF
--- a/src/main/java/me/karakelley/http/filesystem/PublicDirectory.java
+++ b/src/main/java/me/karakelley/http/filesystem/PublicDirectory.java
@@ -34,18 +34,14 @@ public class PublicDirectory {
   }
 
   public List<File> getDirectoriesAndFiles(String requestedResource) {
-    File directory = getPath(requestedResource).toFile();
-    if (resourceExists(requestedResource)) {
-      return Arrays.stream(directory.listFiles())
-              .filter(file -> file.isDirectory() || file.isFile())
-              .collect(Collectors.toList());
-    } else {
-      return new ArrayList<>();
-    }
+    File directory = normalizeFullPath(requestedResource);
+    return Arrays.stream(directory.listFiles())
+            .filter(file -> file.isDirectory() || file.isFile())
+            .collect(Collectors.toList());
   }
 
   public boolean resourceExists(String path) {
-    return fileFinderCache.resourceExists(documentRoot+path);
+    return isValidPath(path) && fileFinderCache.resourceExists(normalizedDocumentRoot() + path);
   }
 
   public String relativePath(File file) {
@@ -77,7 +73,7 @@ public class PublicDirectory {
     new File(newFile.getParentFile().getAbsolutePath()).mkdirs();
 
     if (contents != null) {
-      try(FileOutputStream outputStream = new FileOutputStream(newFile)) {
+      try (FileOutputStream outputStream = new FileOutputStream(newFile)) {
         outputStream.write(contents);
       } catch (IOException e) {
         throw new RuntimeException(e);
@@ -97,4 +93,28 @@ public class PublicDirectory {
       throw new RuntimeException(e);
     }
   }
+
+  private boolean isValidPath(String requestedPath) {
+    String canonicalPath = normalizeFullPath(requestedPath).toString();
+
+    try {
+      return canonicalPath.startsWith(normalizedDocumentRoot().toFile().getCanonicalFile().toString());
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private File normalizeFullPath(String requestPath) {
+    try {
+      return Paths.get(normalizedDocumentRoot() + requestPath).toFile().getCanonicalFile();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private Path normalizedDocumentRoot() {
+    return documentRoot.toPath().normalize();
+  }
+
 }
+
--- a/src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
+++ b/src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
@@ -28,7 +28,9 @@ public class StaticFilesHandler implements Handler {
     if (publicDirectory.resourceExists(requestedResource)) {
       response = new Ok();
       return serveResponse(response, requestedResource);
-    } else return new NotFound();
+    } else {
+      return new NotFound();
+    }
   }
 
   private Response serveResponse(Response response, String requestedResource) {
--- a/src/test/java/me/karakelley/http/publicdirectory/PublicDirectoryTest.java
+++ b/src/test/java/me/karakelley/http/publicdirectory/PublicDirectoryTest.java
@@ -12,7 +12,6 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
 
@@ -77,13 +76,6 @@ class PublicDirectoryTest {
   }
 
   @Test
-  void testNoFilesInDirectory() {
-    PublicDirectory publicDirectory = PublicDirectory.create("/", new FileFinderCache(new RealFileFinder()));
-
-    assertEquals(new ArrayList<>(), publicDirectory.getDirectoriesAndFiles("/somewhere"));
-  }
-
-  @Test
   void testGetMimeType() {
     TempFilesHelper.withTempDirectory(directory -> {
       PublicDirectory publicDirectory = PublicDirectory.create(directory.toString(), new FileFinderCache(new RealFileFinder()));
@@ -142,7 +134,6 @@ class PublicDirectoryTest {
       } catch (IOException e) {
         e.printStackTrace();
       }
-
       assertTrue(publicDirectory.resourceExists("/testing123.txt"));
     });
   }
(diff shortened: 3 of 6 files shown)


src/main/java/me/karakelley/http/exceptions/AttemptedDirectoryTraversalException.java
```diff
@@ -0,0 +1,7 @@
+package me.karakelley.http.exceptions;
+
+public class AttemptedDirectoryTraversalException extends Exception {
```
mikeknep: Any particular reason this one is checked, and not a `RuntimeException` like your others? (You recently made all your others unchecked, so just wondering why this one is different.)

src/main/java/me/karakelley/http/server/HttpRequestReader.java
```diff
@@ -78,8 +81,13 @@ private String readRequestLineAndHeaders() throws IOException {
     return buffer.toByteArray();
   }
 
-  private String validPath(String path) {
-    if (path == null) throw new InvalidRequestException("Invalid Path!");
+  private String validPath(String path) throws AttemptedDirectoryTraversalException {
+    Path userPath = Paths.get(path);
+
+    Path normalizedPath = userPath.normalize();
+    if (normalizedPath != userPath || userPath.startsWith("..")) {
```
mikeknep: 1. Can you walk me through how `normalize` and this comparison works? I'm guessing `normalize` does something like:
```
Paths.get("/Users/mikeknepper/projects/server/../../projects").normalize()
# => "/Users/mikeknepper/projects" (as a Path)
```
Does the comparison then check that they end up in the same place?

2. Could someone get around this check by requesting `/someDirectory/../../.ssh` ? (Do you possibly need to check if `normalizedPath` starts with dots instead of `userPath`?)

mikeknep (commented):
A couple questions, but overall this PR is really nice—it's tightly focused on a specific bug making it easily digestible, and it provides tests to ensure the bug does not reoccur after potential future refactoring. I would prefer seeing the tests and the production code together in a single commit (I might argue this whole PR could be a single commit) but in general very nice job!

src/main/java/me/karakelley/http/exceptions/AttemptedDirectoryTraversalException.java
parkerpersaud: I wanted to be more explicit with this particular exception. I'm not sure if that's a valid enough answer but I'm still going back and forth on pros/cons on checked vs unchecked.

src/main/java/me/karakelley/http/server/HttpRequestReader.java
parkerpersaud: Yes, that's how `normalize` works - So once the path has been normalized and it's different that what you requested then it'll throw an `AttemptedDirectoryTraversalException`. 

So if you requested `/somedirectory/../../.ssh` it'll be normalized to /somedirectory/.ssh. Then we check to see if `/somedirectory/../../.ssh` is equal to ` /somedirectory/.ssh`. They're not equal so exception is thrown. 

I believe checking either `userPath` or `normalizedPath` for preceding dots will the same since `normalize()` doesn't remove any preceding dots however it does seem more readable to check the `normalizedPath` instead.

src/main/java/me/karakelley/http/filesystem/PublicDirectory.java
```diff
@@ -35,17 +35,13 @@ private PublicDirectory(String path, FileFinder fileFinderCache) {
...
   public List<File> getDirectoriesAndFiles(String requestedResource) {
     File directory = getPath(requestedResource).toFile();
-    if (resourceExists(requestedResource)) {
-      return Arrays.stream(directory.listFiles())
-              .filter(file -> file.isDirectory() || file.isFile())
-              .collect(Collectors.toList());
-    } else {
-      return new ArrayList<>();
-    }
+    return Arrays.stream(directory.listFiles())
```
mikeknep: Always give props for removing an `if/else`!

src/main/java/me/karakelley/http/filesystem/PublicDirectory.java
```diff
@@ -91,10 +87,22 @@ public Path getPath(String requestedResource) {
 
   public void updateFileContents(String path, byte[] contents) {
     File newFile = getPath(path).toFile();
-    try (FileOutputStream outputStream = new FileOutputStream(newFile, false)) {
+    try(FileOutputStream outputStream = new FileOutputStream(newFile, false)) {
```
mikeknep: Ultra nitpick: your diff adds a space after one `try` (76) and removes a space for another `try` (here).

src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
```diff
@@ -28,7 +28,9 @@ public Response respond(Request request) {
     if (publicDirectory.resourceExists(requestedResource)) {
       response = new Ok();
       return serveResponse(response, requestedResource);
-    } else return new NotFound();
+    } else {
+      return new NotFound();
```
mikeknep, line 32: Thank you! Again it's a nit, but this style is much more idiomatic.

src/test/java/me/karakelley/http/publicdirectory/PublicDirectoryTest.java
```diff
@@ -76,13 +75,6 @@ void testResourceDoesntExist() {
     assertFalse(publicDirectory.resourceExists("/someBadPath"));
   }
 
-  @Test
-  void testNoFilesInDirectory() {
```
mikeknep, line 80: Out of curiosity, why did you remove this test?

src/main/java/me/karakelley/http/filesystem/PublicDirectory.java
```diff
@@ -91,10 +87,22 @@ public Path getPath(String requestedResource) {
...
       outputStream.write(contents);
     } catch (IOException e) {
       throw new RuntimeException(e);
     }
   }
+
+  public boolean isValidPath(String requestedPath) {
```
mikeknep: Does any external client call this method? Does it need to be public? Or is it only public for purposes of testing?

src/test/java/me/karakelley/http/publicdirectory/PublicDirectoryTest.java
```diff
@@ -197,4 +187,19 @@ void testUpdateContentsOfFile() {
...
+  @Test
+  void testValidPath() {
+    TempFilesHelper.withTempDirectory(directory ->  {
+      PublicDirectory publicDirectory = PublicDirectory.create(directory.toString(), new FileFinderCache(new RealFileFinder()));
+      assertEquals(true, publicDirectory.isValidPath("testing/../testing"));
+    });
+  }
+
+  @Test
+  void testInvalidPath() {
```
mikeknep, line 199: Follow-up on the public-vs-private `isValidPath` comment:
On the one hand, I like these two tests here because they help document the behavior of `isValidPath` method, which I personally find a little hard to follow (but probably mainly because it uses a lot of std lib methods with which I'm not super familiar). On the other hand, the 404-on-attempted-dir-traversal test does cover the public-facing behavior at a nice, high level. Maybe there's a compromise, keeping the 404 test and replacing these with something just a little higher up in `PublicDirectory`?

src/test/java/me/karakelley/http/publicdirectory/PublicDirectoryTest.java
parkerpersaud: I removed this particular test because logically you'd never get to this point in the program if the resource didn't exist (the naming was poor and doesn't express this) now that the validation is all done at a higher level. A test similar to this does exist at a higher level as well.

src/test/java/me/karakelley/http/publicdirectory/PublicDirectoryTest.java
parkerpersaud: The method should be private. I initially had a different implementation and when I changed it to be called within `resourceExists()` instead I forgot to change it to private. I could easily test this through `resourceExists()` which is probably more appropriate entry point for testing it.

mikeknep (approved):

src/main/java/me/karakelley/http/filesystem/PublicDirectory.java
```diff
@@ -97,4 +93,24 @@ public void updateFileContents(String path, byte[] contents) {
       throw new RuntimeException(e);
     }
   }
+
+  private boolean isValidPath(String requestedPath) {
+    File canonicalPath = normalizeFullPath(requestedPath);
+
+    return canonicalPath.toString().contains(normalizedDocumentRoot().toString());
```
aaronlahey: It feels strange to be checking if a path is valid by doing a string comparison. Can you use [this](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#startsWith(java.nio.file.Path))?

src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
aaronlahey: 💯

aaronlahey (approved):
Approved with my one nitpick addressed.

src/main/java/me/karakelley/http/filesystem/PublicDirectory.java
parkerpersaud: So if I pass in the directory `./thispath/somewhere` the canonical path is something like `/Users/karakelley/code/thispath/somewhere` and `normalizedDocumentRoot` would be `/thispath/somewhere` which means the canonical path would never start with `normalizedDocumentRoot` (because its missing part of the full path). I really just want to see if the full document root is included within the requested path. I can see if there might be another alternative to string comparison but so far not seeing anything in the `Path` interface.

PR #19, opened 2018-04-12
Remove line reader

parkerpersaud:
Remove LineReader and single test that relied on the LineReader. That test was moved to HttpRequestReader. 

LineReader is no longer being used with the introduction of reading raw bytes.


mikeknep (approved):
The new test here is sweet. And red diffs are even better than green ones

src/test/java/me/karakelley/http/server/HttpRequestReaderTest.java
```diff
@@ -85,9 +91,37 @@ void testWhenABodyIsPresent() {
     assertEquals("hey", new String(request.getBody()));
   }
 
+  @Test
+  void testThatIOExceptionIsCaughtAndLogged() {
```
aaronlahey: Good test name.

src/test/java/me/karakelley/http/server/HttpRequestReaderTest.java
```diff
@@ -85,9 +91,37 @@ void testWhenABodyIsPresent() {
...
+    });
+  }
+
   private Request parse(String request) {
     ByteArrayInputStream inputStream = new ByteArrayInputStream(request.getBytes());
     HttpRequestReader requestReader = new HttpRequestReader(inputStream);
     return requestReader.read(0);
   }
+
+  private void withAppender(Consumer<Logger> loggerConsumer) {
```
aaronlahey: I like that you're managing your global state by using a lambda instead of polluting your test with `try`/`finally`. I'm not sure why you need a `Consumer<Logger>` as opposed to a `Runnable`... you don't seem to use the `logger` argument above.

I feel like it could be cleaner if you implemented something like this...

```
@Test
void testThatIOExceptionIsCaughtAndLogged() {
  // Not sure if `LogEvent` is the right type here, but you get the idea...
  List<LogEvent> logEvents = withCapturedLogging(() -> {
    ByteArrayInputStream inputStream = new ByteArrayInputStream("GET / HTTP\r\n\r\n".getBytes());
    HttpRequestReader requestReader = new HttpRequestReader(inputStream);
    try {
      inputStream.close();
    } catch (IOException e) {
      e.printStackTrace();
    }
  });

  assertTrue(logEvents.contains("java.io.IOException: Stream closed"));
}

File: src/test/java/me/karakelley/http/server/HttpRequestReaderTest.java [hunk 3]
```diff
@@ -85,9 +91,37 @@ void testWhenABodyIsPresent() {
...
 
+  @Test
+  void testThatIOExceptionIsCaughtAndLogged() {
+    withAppender(logger -> {
+      ByteArrayInputStream inputStream = new ByteArrayInputStream("GET / HTTP\r\n\r\n".getBytes());
+      HttpRequestReader requestReader = new HttpRequestReader(inputStream);
+      try {
+        inputStream.close();
+      } catch (IOException e) {
+        e.printStackTrace();
```
aaronlahey: If you're in a junit test and you're capturing an exception you don't expect to be thrown it's probably a better idea to do something like this...

```
try {
  // do the thing...
} catch (SomeException e) {
  fail(e);
}
```

You'll still get the error message printed via Junit, but your test will fail (which is better than silently passing for the wrong reason)

parkerpersaud: Oh! Cool! I had seen that before but never understood why someone would want their test to fail - seeing it in this context makes much more sense.

src/test/java/me/karakelley/http/server/HttpRequestReaderTest.java
parkerpersaud: Im still shakey on lambdas. I've watched 2 tutorials this morning and read a few more posts and I just can't seem to write them from scratch myself. Most of the things I read/watch say you need to implement an interface which I know I don't need to do in this case. Anyway, if you have time to maybe pair on this today or walk me through lambdas that'd be awesome.

PR #20, opened 2018-04-12
Create Factory to create/configure Application with or without a publ…

parkerpersaud:
Create factory to handle setting up an Application instance with or without a public directory

### DIFF
--- a/src/main/java/me/karakelley/http/ApplicationFactory.java
+++ b/src/main/java/me/karakelley/http/ApplicationFactory.java
@@ -1,0 +1,19 @@
+package me.karakelley.http;
+
+import me.karakelley.http.filesystem.PublicDirectory;
+import me.karakelley.http.handlers.Application;
+import me.karakelley.http.handlers.Handler;
+
+import java.util.Map;
+
+public class ApplicationFactory {
+
+  public Handler create(Map<String, String> argsHash) {
+
+    if (argsHash.containsKey("directory")) {
+      return new Application(PublicDirectory.create(argsHash.get("directory")));
+    } else {
+      return new Application();
+    }
+  }
+}
--- a/src/main/java/me/karakelley/http/Main.java
+++ b/src/main/java/me/karakelley/http/Main.java
@@ -1,31 +1,19 @@
 package me.karakelley.http;
 
-import me.karakelley.http.filesystem.PublicDirectory;
-import me.karakelley.http.handlers.Application;
 import me.karakelley.http.server.*;
 import me.karakelley.http.utility.CommandLineArguments;
 import me.karakelley.http.utility.SystemExit;
 
 import java.util.Map;
-import java.util.function.BiConsumer;
 
 class Main {
   public static void main(String[] args) {
-    main(args, (argsHash1, serverConfiguration1) -> {
-      if (!argsHash1.containsValue("nodirectory")) {
-        serverConfiguration1.setHandler(new Application(PublicDirectory.create(argsHash1.get("directory"))));
-      } else serverConfiguration1.setHandler(new Application());
-    });
-  }
-
-  public static void main(String[] args, BiConsumer<Map<String, String>, ServerConfiguration> configurationBiConsumer) {
-    Map<String, String> argsHash = CommandLineArguments.parse(args, new SystemExit());
+    Map<String, String> argsHash = new CommandLineArguments().parse(args, new SystemExit());
     ServerConfiguration serverConfiguration = new ServerConfiguration();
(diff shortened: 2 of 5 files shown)


src/main/java/me/karakelley/http/ApplicationFactory.java
```diff
@@ -0,0 +1,19 @@
+package me.karakelley.http;
+
+import me.karakelley.http.filesystem.FileFinderCache;
+import me.karakelley.http.filesystem.PublicDirectory;
+import me.karakelley.http.filesystem.RealFileFinder;
+import me.karakelley.http.handlers.Application;
+import me.karakelley.http.handlers.Handler;
+
+public class ApplicationFactory {
+  private final String NO_DIRECTORY = "nodirectory";
```
mikeknep: Hmm. This is a second private constant that has to magically match `CommandLineArguments#DEFAULT_PATH`...

src/main/java/me/karakelley/http/Main.java
```diff
@@ -1,33 +1,19 @@
...
 class Main {
   public static void main(String[] args) {
-    main(args, (argsHash1, serverConfiguration1) -> {
-      if (!argsHash1.containsValue("nodirectory")) {
-        serverConfiguration1.setHandler(new Application(PublicDirectory.create(argsHash1.get("directory"), new FileFinderCache(new RealFileFinder()))));
-      } else serverConfiguration1.setHandler(new Application());
-    });
-  }
-
-  public static void main(String[] args, BiConsumer<Map<String, String>, ServerConfiguration> configurationBiConsumer) {
```
mikeknep, line 21: Thank goodness!

src/test/java/me/karakelley/http/server/HttpServerTest.java
```diff
@@ -29,7 +26,7 @@ void testSendsResponseGivenValidRequest() throws Exception {
     ClientHelper client = new ClientHelper();
     ServerConfiguration config = new ServerConfiguration();
     config.setPort("0");
-    config.setHandler(new Application());
+    config.setHandler(new ApplicationFactory().create("nodirectory"));
```
mikeknep: And, here is a third place the specific string `"nodirectory"` is used...

src/main/java/me/karakelley/http/Main.java
```diff
@@ -1,33 +1,19 @@
...
     Map<String, String> argsHash = CommandLineArguments.parse(args, new SystemExit());
     ServerConfiguration serverConfiguration = new ServerConfiguration();
     serverConfiguration.setPort(argsHash.get("port"));
-
-    configurationBiConsumer.accept(argsHash, serverConfiguration);
-
+    serverConfiguration.setHandler(new ApplicationFactory().create(argsHash.get("directory")));
```
aaronlahey: What are the benefits of a separate class with instance methods vs adding a `static` method on `Application`?

src/test/java/me/karakelley/http/server/HttpServerTest.java
```diff
@@ -45,7 +42,7 @@ void testSendsResponseWithHeadersGivenValidRequest() throws Exception {
...
```
aaronlahey: I feel like there have been multiple PRs now where every test has needed the same small tweak. Is there any logic that can be extracted into `private` methods?

src/main/java/me/karakelley/http/ApplicationFactory.java
aaronlahey: Yeah, I'd say giving special meaning to certain string values like this is almost always a bad idea. Even in languages where it's common (JavaScript) I still try to avoid this. What about just having two separate overloads? `create(publicDirectory)` and `create()`, or just two separate methods? `createWithPublicDirectory`/`createWithoutPublicDirectory`.

aaronlahey:
I'd say this is an improvement over what was there. Getting the logic out of `main` is a good thing so I like where this PR is headed.

src/test/java/me/karakelley/http/server/HttpServerTest.java
parkerpersaud: I've been wanting to do that for a while but I thought moving all the object instantiation to each individual test was to help with visibility? At what point is it ok to extract some duplication to a method? Is visibility preferred over duplication?

src/main/java/me/karakelley/http/ApplicationFactory.java
parkerpersaud: But at some point I need to check if a directory is present within the args, right?

PR #23, opened 2018-04-15
Refactor validation to standalone class and add basic authorization

parkerpersaud:
https://trello.com/c/skDDioAf/104-525-basic-auth


src/main/java/me/karakelley/http/responses/Unauthorized.java
```diff
@@ -0,0 +1,19 @@
...
+
+public class Unauthorized extends Response {
+
+  public String getStatus() {
+    return "401 Unauthorized";
+  }
+
+  public Map<String, String> getHeaders() {
+    Map<String, String> headers = new HashMap<>();
+    headers.put("WWW-Authenticate", "Basic realm=My Server");
```
mikeknep: Nice job consistently getting your responses up to spec. You've been nailing the response codes, and now this is one more thing that could easily be forgotten.

src/main/java/me/karakelley/http/server/RequestReaderFactory.java
```diff
@@ -1,9 +1,12 @@
...
 
+import me.karakelley.http.authorization.Authorizer;
+import me.karakelley.http.RequestValidator;
+
 import java.io.InputStream;
 
 public class RequestReaderFactory {
   public HttpRequestReader getReader(InputStream inputStream) {
-    return new HttpRequestReader(inputStream);
+    return new HttpRequestReader(inputStream, new RequestValidator(new Authorizer("admin", "chicago32")));
```
mikeknep: I definitely like that you're injecting the required username and password. That said, I wonder if you could bring that injection up even higher. If I was deploying your server in dev, staging, and production environments and wanted different creds for each environment to keep goofballs out of prod, how might I do that? (Note: don't go overboard with this—I don't need a yaml config file or environment variables... but I think there's probably some compromise between this and that)

src/main/java/me/karakelley/http/HttpMethod.java
```diff
@@ -26,4 +29,12 @@ public static HttpMethod fromString(String s) {
         throw new InvalidRequestException("ouch");
     }
   }
+
+  public static List<HttpMethod> modifiableMethods() {
```
mikeknep: This could probably be a constant?

src/main/java/me/karakelley/http/RequestValidator.java
```diff
@@ -0,0 +1,42 @@
+package me.karakelley.http;
+
+import me.karakelley.http.authorization.Authorizer;
+import me.karakelley.http.exceptions.InvalidRequestException;
+import me.karakelley.http.exceptions.UnauthorizedException;
+import static me.karakelley.http.HttpMethod.modifiableMethods;
```
mikeknep: @aaronlahey would know better than me—is this cool or frowned upon in Java?

src/main/java/me/karakelley/http/server/RequestReaderFactory.java
parkerpersaud: Im having a difficult time thinking of a way to do this without using env variables or a config file. I could pass the values in on the command line but I don't particularly like that solution either.

src/main/java/me/karakelley/http/RequestValidator.java
```diff
@@ -0,0 +1,41 @@
...
+    return path;
+  }
+
+  public String validProtocol(String protocol) {
+    if (!protocol.equals("HTTP/1.1")) throw new InvalidRequestException("Invalid Protocol!");
+    return protocol;
+  }
+
+  public boolean requestIsAuthorized(HttpMethod method, Map<String, String> headers) {
+    String credentials = headers.getOrDefault("Authorization", null);
```
mikeknep: nit pick: this could be inside the `if` statement, since you don't need to check this for `GET` requests.

mikeknep: Additionally, there's an opportunity to simplify the authorizer here. This `getOrDefault("Authorization", null)` means that `String credentials` could be `null`, which you're then passing to the authorizer, which has to do a null check. There are a couple alternative options. You could use Optionals in some way (not sure what Java gives you for this, but fyi there are some languages where accessing a key from a hash will *always* return an Optional). If you don't want to get that intense, you could modify `checkAuthorization` to do something like:
```
if (credentials != null && authorizer.isAuthorized(credentials)) {...}
```

src/main/java/me/karakelley/http/authorization/Authorizer.java
```diff
@@ -0,0 +1,31 @@
...
+    this.user = user;
+    this.password = password;
+  }
+
+  public boolean isAuthorized(String credentials) {
+    return encodeAuthoriziation().equals(parseRequestAuthorization(credentials));
+  }
+
+  private String parseRequestAuthorization(String credentials) {
+    if (credentials == null) {
```
mikeknep: See comment above

src/test/java/me/karakelley/http/server/HttpRequestReaderTest.java
```diff
@@ -42,7 +42,7 @@ void testRequestHasGetMethod() {
   void testRequestHasPostMethod() {
     Request request = parse(basicPostRequest());
 
-    assertEquals(HttpMethod.POST, request.getMethod());
+   assertEquals(HttpMethod.POST, request.getMethod());
```
mikeknep: This space shouldn't have been deleted

src/test/java/me/karakelley/http/server/HttpRequestReaderTest.java
```diff
@@ -81,10 +81,11 @@ void testInvalidHttpProtocol() {
 
   @Test
   void testRequestHasHeaders() {
-    requestString = requestBuilder.setMethod("POST").setPath("/hey.txt").setHeader("Content-Type", "text/plain").build();
+    requestString = requestBuilder.setMethod("POST").setPath("/hey.txt").setHeader("Authorization", "Basic YWRtaW46Y2hpY2FnbzMy").setHeader("Content-Type", "text/plain").build();
     Request request = parse(requestString);
     Map<String, String> headers = new HashMap<>();
     headers.put("Content-Type", "text/plain");
+    headers.put("Authorization", "Basic YWRtaW46Y2hpY2FnbzMy");
```
mikeknep: Why add this header to this test? You already have a test below for multiple headers

src/test/java/me/karakelley/http/server/HttpServerTest.java
```diff
@@ -99,16 +100,11 @@ void testSends400GivenInvalidProtocol() {
   @Test
   void testPortIsUnavailable() {
     List<String> events = withCapturedLogging(() -> {
-      ServerConfiguration config = new ServerConfiguration();
-      config.setPort("0");
-      Map<String, String> args = new HashMap<>();
-      config.setHandler(new ApplicationFactory().create(args));
-      config.setPort("4000");
-      HttpServer httpServer = new HttpServer(config, new ConnectionHandler(), new RequestReaderFactory());
+      HttpServer httpServer = configureWithNoDirectory("4000");
```
mikeknep, line 104: I like this change. We've talked about the difficulty of finding the "right" balance between explicit verbosity in each test vs. extracting commonality into private functions... it becomes a style and "feels right/wrong" kind of thing, and I think your intuition on it is getting stronger.

src/main/java/me/karakelley/http/handlers/ApplicationAuthorizationFactory.java
```diff
@@ -0,0 +1,28 @@
...
+import me.karakelley.http.server.Handler;
+
+import java.util.Map;
+
+public class ApplicationAuthorizationFactory {
+
+  private Authorization basicAuth;
+  private Authorization alwaysAuthorized;
+
+  public ApplicationAuthorizationFactory(Authorization basicAuth, Authorization alwaysAuthorized) {
```
mikeknep: 1. It's going to be hard to remember the order of these two arguments—"which one comes first? and what determines which one gets used again?" I don't have a great suggestion at the moment... 
2. The class name is better, but something still feels a little strange to me about it. Maybe just a personal style thing. You're creating a handler that combines an application with some form of auth protection/security. It's a subtle, "English-grammar" thing, but would `AuthorizedApplicationFactory` make more sense? (adjective before noun.) For some reason that name also feels to me a little more indicative of the fact that the logic in this class is specific to this application, whereas `ApplicationAuthorizationFactory` sounds like a potentially generic tool for you to build some kind of authorization object that someone/anyone might use for their own application. 🤔 does that make any sense?

PR #24, opened 2018-04-15
Refactor all  file-type handlers

parkerpersaud:
Refactor all the handlers that deal with the file system to be dispatched through StaticFilesHandler


src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
```diff
@@ -1,65 +1,39 @@
...
-    ArrayList<String> resources = new ArrayList<>();
-    for (File file : files) {
-      resources.add(contentPresenter.displayResources(file, publicDirectory));
-    }
-    return resources;
+  private void setupFileStrategies() {
+    fileStrategies.put(HttpMethod.POST, new NewResourceStrategy(publicDirectory));
+    fileStrategies.put(HttpMethod.GET, new RetrieveResourceStrategy(publicDirectory, htmlPresenter));
+    fileStrategies.put(HttpMethod.DELETE, new DeleteResourceStrategy(publicDirectory));
+    fileStrategies.put(HttpMethod.PUT, new UpdateResourceStrategy(publicDirectory));
```
mikeknep: Does this need to happen on every request? Or can this be some static, constant map?

src/main/java/me/karakelley/http/handlers/Application.java
```diff
@@ -17,10 +17,11 @@ public Application() {
...
     this.publicDirectory = publicDirectory;
-    router.route(HttpMethod.POST, new NewResourceHandler(publicDirectory));
-    router.route(HttpMethod.GET, new StaticFilesHandler(publicDirectory, new HtmlPresenter()));
-    router.route(HttpMethod.PUT, new UpdateResourceHandler(publicDirectory));
-    router.route(HttpMethod.DELETE, new DeleteResourceHandler(publicDirectory));
+    Handler staticFileHandler = new StaticFilesHandler(publicDirectory, new HtmlPresenter());
+    router.route(HttpMethod.GET, staticFileHandler);
+    router.route(HttpMethod.POST, staticFileHandler);
+    router.route(HttpMethod.PUT, staticFileHandler);
+    router.route(HttpMethod.DELETE, staticFileHandler);
```
mikeknep: I'm of two minds on this. On one hand, the granularity here is nice: I like how the full spectrum of requests supported by your server is at one "level", so (for example) it's easy to see that your server does not support HEAD/OPTIONS requests, and (as another example) if some other user of your server only wants to support read actions, they can just route in a GET with the StaticFilesHandler and not add POST/PUT/DELETE routes. I could see an argument that this is better than if all the methods were supported within StaticFilesHandler such that someone who only wants read actions would have to recreate the GET portion of that handler without all the rest.

On the other hand, there are other ways someone could get that read-only functionality (some kind of auth filter preventing write requests), and it still feels a little strange to me that, from the client perspective, your generic filesystem-y routes seem to get wired in at the same level of granularity as your domain-specific routes (i.e. /redirectme, /parse). The method signature (i.e. the lack of a specific path argument) does help, but I keep feeling like I want some sort of API like "on GET /redirectme do this, on GET /parse do that, and on anything else, here's a staticFilesHandler you can use to potentially do something in the public directory you're serving". Does that make sense? I don't think I'm explaining it well. @aaronlahey do you have any thoughts on this?

src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
aaronlahey: If it's safe to share the same instance of a handler across requests (they're stateless) I would put this in a `static` read-only map. You can use a `static` block to set that up...

```
private static final Map<HttpMethod, Handler> fileStrategies;

static {
  HashMap<HttpMethod, Handler> m = new HashMap<>();
  m.put(...);
  fileStrategies = Collections.unmodifiableMap(m);
}
```

src/main/java/me/karakelley/http/handlers/Application.java
aaronlahey: >...and on anything else, here's a staticFilesHandler you can use to potentially do something in the public directory you're serving.

I would agree with this. In fact, that's what's already happening... it just isn't explicit. You're mapping all known HTTP methods and all known paths (by omitting the path) to the same handler.

src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
```diff
@@ -1,65 +1,39 @@
...
-  private final ContentPresenter contentPresenter;
+  private final Map<HttpMethod, Handler> fileStrategies = new HashMap<>();
   private final PublicDirectory publicDirectory;
-  private final String CONTENT_TYPE = "Content-Type";
-  private final String TEXT_HTML = "text/html";
+  private final ContentPresenter htmlPresenter;
 
-  public StaticFilesHandler(PublicDirectory publicDirectory, ContentPresenter contentPresenter) {
-    this.contentPresenter = contentPresenter;
+  public StaticFilesHandler(PublicDirectory publicDirectory, ContentPresenter htmlPresenter) {
```
aaronlahey: Why the variable name change?

src/test/java/me/karakelley/http/handlers/NewResourceStrategyTest.java
```diff
@@ -12,14 +13,14 @@
 
 import static org.junit.jupiter.api.Assertions.*;
 
-class NewResourceHandlerTest {
+class NewResourceStrategyTest {
```
aaronlahey: I think these strategies are probably implementation details. I would test all this behavior through `StaticFilesHandler`. Then you're free to refactor within that boundary. If you decide you want to pull common behavior up into that `StaticFilesHandler` or merge all these into one class you don't want to be breaking all these tests.

aaronlahey: This is probably what I would use that nested class syntax in JUnit 5 for (a nested context for each method). I think this syntax is controversial, but I like it...

```
public class StaticFilesHandlerTest {
  @Nested
  public class when_the_request_is_a_get {
    @Nested
    public class when_the_requested_resource_exists {
      public void sets_the_file_contents_as_the_body() {
        //...
      }

      public void sets_the_content_type_to_application_html_for_files_with_an_html_file_extension() {
        //...
      }
    }
  }
}

File: src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java [hunk 5]
```diff
@@ -1,65 +1,41 @@
...
+  private static Map<HttpMethod, Handler> fileStrategies;
+  private static PublicDirectory publicDirectory = null;
+  private static ContentPresenter htmlPresenter = null;
 
-  public StaticFilesHandler(PublicDirectory publicDirectory, ContentPresenter contentPresenter) {
-    this.contentPresenter = contentPresenter;
+  public StaticFilesHandler(PublicDirectory publicDirectory, ContentPresenter htmlPresenter) {
     this.publicDirectory = publicDirectory;
+    this.htmlPresenter = htmlPresenter;
+    setupFileStrategies();
```
aaronlahey: This would re-instantiate the strategies every time the class is instantiated.

1. Changing `static` level things from the instance level is almost always a bad idea
2. This isn't any more efficient than just having `fileStrategies` be an instance variable.

Try changing this code so the `Map` is only set up one time for the life of the program.

src/test/java/me/karakelley/http/handlers/StaticFilesHandlerTest.java
```diff
@@ -1,59 +1,152 @@
...
-      ContentPresenter contentPresenter = new HtmlPresenter();
-      Handler handler = new StaticFilesHandler(publicDirectory, contentPresenter);
-      Response response = handler.respond(new Request(HttpMethod.GET, "/", "HTTP/1.1", null, null,0));
 
-      assertTrue(new String(response.getBody()).split("", 2).length == 2);
-    });
+  @Nested
+  class WhenRequestIsGet {
+    @Test
+    void testDisplaysFiles() {
```
aaronlahey: Is there a test for when the file isn't there?

src/test/java/me/karakelley/http/handlers/StaticFilesHandlerTest.java
```diff
@@ -1,59 +1,152 @@
...
-      ContentPresenter contentPresenter = new HtmlPresenter();
-      Handler handler = new StaticFilesHandler(publicDirectory, contentPresenter);
-      Response response = handler.respond(new Request(HttpMethod.GET, "/", "HTTP/1.1", null, null, 0));
-
-      assertTrue(new String(response.getBody()).contains("<p><a href=\"/test1.txt\">test1.txt</a></p>"));
-    });
+  @Nested
+  class WhenRequestIsPost {
+    @Test
+    void test201ResponseForPostContainsLocationHeader() {
```
aaronlahey: Is that what this is even testing?

src/test/java/me/karakelley/http/handlers/StaticFilesHandlerTest.java
```diff
@@ -1,59 +1,152 @@
...
+        assertEquals("201 Created", response.getStatus());
+      });
+    }
   }
-}
 
+  @Nested
+  class WhenRequestIsDelete {
+    @Test
+    void testResourceIsDeleted() {
```
aaronlahey: Is there a test for if the file doesn't exist?

src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
parkerpersaud: I don't think that's true. Once the server starts, `Application` is instantiated only once. Upon instantiation it then instantiates a new `StaticFilesHandler` which then instantiates all the strategies (just once) and it never happens again (unless you restart the server obviously).

src/test/java/me/karakelley/http/handlers/StaticFilesHandlerTest.java
parkerpersaud: There is a test for this scenario in `HttpServerTest`

src/test/java/me/karakelley/http/handlers/StaticFilesHandlerTest.java
parkerpersaud: There is a test for this scenario in `HttpServerTest`

src/test/java/me/karakelley/http/handlers/StaticFilesHandlerTest.java
parkerpersaud: Not quite - I will update this to be more expressive

PR #25, opened 2018-04-16
Feat index.html

parkerpersaud:
https://trello.com/c/Dj8GxhGG/105-425-indexhtml

### DIFF
--- a/src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
+++ b/src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
@@ -10,12 +10,15 @@ import me.karakelley.http.server.Handler;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
 
 public class StaticFilesHandler implements Handler {
   private final ContentPresenter contentPresenter;
   private final PublicDirectory publicDirectory;
   private final String CONTENT_TYPE = "Content-Type";
   private final String TEXT_HTML = "text/html";
+  private final String ROOT_INDEX_FILE = "index.html";
 
   public StaticFilesHandler(PublicDirectory publicDirectory, ContentPresenter contentPresenter) {
     this.contentPresenter = contentPresenter;
@@ -42,6 +45,10 @@ public class StaticFilesHandler implements Handler {
   }
 
   private Response serveDirectoryResponse(Response response, String requestedResource) {
+    Optional<String> indexFile = lookForIndexFile(requestedResource);
+    if (indexFile.isPresent()) {
+      return serveFilesResponse(response, "/" + indexFile.get());
+    }
     response.setBody(String.join("", getFilesAndDirectoryLinks(requestedResource)));
     response.setHeaders(CONTENT_TYPE, TEXT_HTML);
     return response;
@@ -60,6 +67,16 @@ public class StaticFilesHandler implements Handler {
       resources.add(contentPresenter.displayResources(file, publicDirectory));
     }
     return resources;
+  }
+
+  private Optional<String> lookForIndexFile(String requestedResource) {
+    List<File> files = publicDirectory.getDirectoriesAndFiles(requestedResource);
+
+    List<String> result = files.stream()
+            .map(File::getName)
+            .filter(fileName -> fileName.equals(ROOT_INDEX_FILE))
+            .collect(Collectors.toList());
+    return result.stream().findFirst();
   }
 }
 
--- a/src/test/java/me/karakelley/http/handlers/StaticFilesHandlerTest.java
+++ b/src/test/java/me/karakelley/http/handlers/StaticFilesHandlerTest.java
@@ -7,6 +7,7 @@ import me.karakelley.http.http.HttpMethod;
 import me.karakelley.http.http.Response;
 import me.karakelley.http.http.Request;
 import me.karakelley.http.server.Handler;
+import me.karakelley.http.server.HttpServer;
 import org.junit.jupiter.api.Test;
 
 import java.nio.file.Path;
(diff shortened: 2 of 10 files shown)


src/test/java/me/karakelley/http/helpers/TempFilesHelper.java
```diff
@@ -31,9 +31,9 @@ public static void withTempDirectory(Consumer<Path> f) {
     }
   }
 
-  public static Path createTempFile(Path directory, String name) {
+  public static Path createTempFile(Path directory, String name, String extension) {
```
mikeknep: Any particular reason you made the extension a third argument instead of just appending it to the `name` argument (i.e. `TempFilesHelper.createTempFile(directory, "/test1.txt")`)?
OR, you could overload `createTempFile` to default to `.txt` when only two args are provided, or provide a third extension argument.

It just seems a little strange to have to specify `".txt"` in all these tests that don't seem to actually use/care about it.

src/test/java/me/karakelley/http/handlers/StaticFilesHandlerTest.java
```diff
@@ -17,8 +17,8 @@
   @Test
   void testGetDisplayFiles() {
     TempFilesHelper.withTempDirectory(directory -> {
-      Path fileOne = TempFilesHelper.createTempFile(directory, "/test1");
-      Path fileTwo = TempFilesHelper.createTempFile(directory, "/test2");
+      Path fileOne = TempFilesHelper.createTempFile(directory, "/test1", ".txt");
```
aaronlahey: Why did you need to add file extensions in all these tests?

src/test/java/me/karakelley/http/server/HttpServerTest.java
```diff
@@ -485,6 +484,31 @@ void test404ForGetRequestAfterResourceDeleted() {
     });
   }
 
+  @Test
+  void testServesHtmlFileForDirectoryWhenPresent() {
```
aaronlahey, line 369: If your computer didn't have copy/paste what do you think the tests in this file would look like? There's so much boilerplate in each one it's hard to tell what's important.

parkerpersaud: I agree - I am in the process of refactoring the tests in another PR

aaronlahey: 👍 Nice job not muddying this PR with those changes.

src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
```diff
@@ -61,5 +67,17 @@ private Response serveFilesResponse(Response response, String requestedResource)
...
+  private List<String> lookForIndexFile(String requestedResource) {
+    List<File> files = publicDirectory.getDirectoriesAndFiles(requestedResource);
+    return files.stream()
+            .map(File::getName)
+            .filter(fileName -> fileName.equals(ROOT_INDEX_FILE))
+            .collect(Collectors.toList());
+  }
+
+  private boolean indexFileIsPresent(List<String> files) {
+    return files.size() > 0;
```
aaronlahey: This is a sure sign you want `Optional` rather than a `List`. An optional is just a way to represent a value that is either there or not...

```
Optional.ofNullable(42); //=> Some(42)
Optional.ofNullable(null); //=> None
```

You can also think of an optional as a `List` that can only ever be empty or have one element.

mikeknep: I know you don't have much experience with `Optional`, so I want to clarify this comment a bit. `ofNullable` is a constructor—it's a way of "wrapping" a value inside an `Optional`. A different example (pseudo coding a bit here between Java and Scala) could be a function like `parseInt` with a type signature `String -> Optional<Integer>`. It could operate like this:
```
x = "23"
y = "hello"
xAsInt = parseInt(x) //=> Some(23)
yAsInt = parseInt(y) //=> None
```
What's great about this is that anything calling `parseInt` is forced, by the compiler, to address the possibility of `parseInt` not having returned a valid integer. This is because the _types_ of `xAsInt`  and `yAsInt` are not `Integer`/`null`, but instead both are `Optional<Integer>` (with one "under the hood" being a `Some<Integer>` and the other being `None`).

What do I mean by "forced to address the possibility"? Well, operations like `+` and `-` are not defined on `Optional<Integer>`—instead, you need to "unwrap" them:
```
def maybeAddFive(z: Optional<Integer>): Integer = {
  z match {
    case Some(a) => a + 5
    case None => // throw an exception, use some default, whatever... but you have to choose something!
  }
}
```

parkerpersaud: I don't understand how this can be used in place of what I have though. I'm checking to see if a file (index.html) was found, if so do this, otherwise do this instead.

src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
```diff
@@ -42,6 +45,10 @@ private Response serveResponse(Response response, String requestedResource) {
   }
 
   private Response serveDirectoryResponse(Response response, String requestedResource) {
+    Optional<String> indexFile = lookForIndexFile(requestedResource);
```
aaronlahey, line 48: 🔥 🔥 🔥

src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
```diff
@@ -42,6 +45,10 @@ private Response serveResponse(Response response, String requestedResource) {
...
+    if (indexFile.isPresent()) {
+      return serveFilesResponse(response, "/" + indexFile.get());
```
aaronlahey, line 50: Small nitpick about `Optionals`. If I can avoid it I try to always avoid `get()`. That will thrown an exception if you haven't checked `isPresent` first. It doesn't feel like you've bought much over `null` in that case. If you always avoid `get()` you're forced to deal with the issue by using things like `map` and `getOrElse`.

src/test/java/me/karakelley/http/publicdirectory/PublicDirectoryTest.java
```diff
@@ -26,7 +26,7 @@ void testMissingDirectory() {
   @Test
   void testPathPointsToAFile() {
     TempFilesHelper.withTempDirectory(directory -> {
-      Path file = TempFilesHelper.createTempFile(directory, "/test1");
+      Path file = TempFilesHelper.createTempFile(directory, "/test1.txt");
```
aaronlahey, line 29: Why were all these changes necessary, again?

aaronlahey (approved):
Approved with the comment about adding `.txt` to everything addressed.

src/test/java/me/karakelley/http/publicdirectory/PublicDirectoryTest.java
parkerpersaud: Because before I had hardcoded `.txt` in the `TempFilesHelper` but in order to create other type of files, i.e. html files which is necessary for this story, I needed to change the implementation to be more dynamic

PR #26, opened 2018-04-17
Feat query params

parkerpersaud:
https://trello.com/c/xZIuifdh/106-600-query-params


src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
```diff
@@ -53,13 +53,17 @@ private Response serveFilesResponse(Response response, String requestedResource)
...
     ArrayList<String> resources = new ArrayList<>();
     for (File file : files) {
-      resources.add(contentPresenter.displayResources(file, publicDirectory));
+      resources.add(createLinks(file));
     }
-    return resources;
+    return contentPresenter.displayResources(String.join("", resources));
+  }
+
+  private String createLinks(File fileName) {
```
mikeknep: Why did this leave the `HtmlPresenter`? Seems pretty HTML-specific to me...

src/main/java/me/karakelley/http/UriQueryParser.java
```diff
@@ -0,0 +1,45 @@
...
+            .collect(Collectors.toMap(pair -> decode(parsePair(pair, 0)),
+                    pair -> Arrays.asList(decode(parsePair(pair, 1)))));
+
+  }
+
+  private static String getQueryString(String requestPath) {
+    try {
+      return URI.create(requestPath).getRawQuery();
+    } catch (Exception e) {
+      throw new InvalidRequestException("Invalid");
```
mikeknep: 1. Are you catching Exception because of `URI.create(requestPath)` or `.getRawQuery()`? If the former, I'm wondering if there's a way you could verify the requestPath is valid ahead of time so that this can focus exclusively on the queryParams?
2. If there's a way to bomb out somewhere else, it'd be nice if this could return an empty string so that you can avoid the `if (queries == null || queries.isEmpty())` check in `parse`. (I'm pretty sure you could just pass `""` to `Arrays.stream` with all the subsequent chained methods and it'll just be a no-op... but maybe I'm wrong about that?)

parkerpersaud: The reason I throw the exception there (or bomb out) is if the query string has malformed Syntax then a `UriSyntaxException` is thrown. But I realized I can get rid of the `isEmpty()` check since `getRawQuery()` returns `null` vs an empty string.

src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
parkerpersaud: I went back and forth on this. This is something I struggle with sometimes when I have a class implementing an interface like `ContentPresenter` and I need a new specific function for an implementing class.  It doesn't make sense for `ContentPresenter` to have a function `createLinks` because that wouldn't make sense if other content types were introduced like JSON. But it also feels a little weird for it to be here considering there's an `HtmlPresenter` class.

src/main/java/me/karakelley/http/Request.java
```diff
@@ -27,6 +27,10 @@ public String getPath() {
     return path;
   }
 
+  public String getPathWithoutQueryParams() {
```
aaronlahey: I would be OK just having a `getPath` that returns the path without the params and ` getQueryParams` method. A URI is this...

```
scheme:[//[user[:password]@]host[:port]][/path][?query][#fragment]
```

The path is just a small part of it. It's already `WithoutQueryParams` by definition.

src/main/java/me/karakelley/http/handlers/QueryParametersHandler.java
```diff
@@ -0,0 +1,50 @@
...
+      return serveQueryResponse(response, request);
+    } catch (InvalidRequestException e) {
+      return new BadRequest();
+    }
+  }
+
+  private Response serveQueryResponse(Response response, Request request) {
+    Map<String, List<String>> query = UriQueryParser.parse(request.getPath());
+    if (!query.isEmpty()) {
+      response.setBody(contentPresenter.displayResources(createDescriptionList(query)));
```
aaronlahey: Using `ContentPresenter` here feels wrong. It feels like what you need here and what you need in static files handler are likely to change for completely different reasons. My gut says that `ContentPresenter` is too vague of a name and it leads one to do this.

src/main/java/me/karakelley/http/handlers/QueryParametersHandler.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+    response.setHeaders(CONTENT_TYPE, TEXT_HTML);
+    return response;
+  }
+
+  private String createDescriptionList(Map<String, List<String>> query) {
+    StringBuilder paramsTable = new StringBuilder();
+    paramsTable.append("<dl>");
+    for (Map.Entry<String, List<String>> entry : query.entrySet()) {
+      paramsTable.append("<dt>" + entry.getKey() + "</dt><dd>" + entry.getValue().get(0) + "</dd>");
```
aaronlahey: Why render some html here and some using the interface formerly known as `ContentPresenter`?

src/main/java/me/karakelley/http/handlers/StaticFilesHandler.java
aaronlahey: Let's chat about this tomorrow (Friday).

src/test/java/me/karakelley/http/handlers/QueryParametersHandlerTest.java
```diff
@@ -0,0 +1,49 @@
...
+  private QueryParametersHandler handler;
+
+  @BeforeEach
+  void setUp() {
+    handler = new QueryParametersHandler(new HtmlPresenter());
+  }
+
+  @Test
+  void testParsesSimpleQuery() {
+    Response response = handler.respond(new Request(HttpMethod.GET, "/parse?hello=world", "HTTP/1.1", null, "".getBytes(), 0));
```
aaronlahey: Can you use the `Builder` here?

src/test/java/me/karakelley/http/handlers/QueryParametersHandlerTest.java
```diff
@@ -0,0 +1,49 @@
...
+    assertTrue(new String(response.getBody()).contains("<dt>hello</dt><dd>world</dd>"));
+  }
+
+  @Test
+  void testParsesQueryWithMultipleParameters() {
+    Response response = handler.respond(new Request(HttpMethod.GET, "/parse?hello=world&hola=mundo,hey", "HTTP/1.1", null, "".getBytes(), 0));
```
aaronlahey: I don't think params should be part of the path. It should be its own concept.

PR #29, opened 2018-04-17
Refactor packages

parkerpersaud:


aaronlahey:
1. I feel like contentpresenter should depend on handlers.
2. I'm not sure about a separate exceptions package. What makes exceptions different than your data objects like HttpMethod or Request/Reponse? I always have a weird feeling in my gut when I see code grouped for code-related reasons (exceptions, enums, constants, interfaces). Ideally code should be grouped for domain-related reasons (http, server, etc).

aaronlahey (approved):

PR #32, opened 2018-04-18
Update response for PUT to return 204

parkerpersaud:


aaronlahey:
Why?

aaronlahey (approved):

parkerpersaud:
200 is an appropriate response but 204 is even more appropriate because it means `The server has successfully fulfilled the request and that there is no additional content to send in the response payload body.`

